PR #4, opened 2023-05-21
Create rust-clippy.yml

hikalium:
Add rust-clippy.yml to enable rust clippy workflow


hikalium:
I created the workflow via GitHub web UI, but it looks like the recipe is obsolete, so I'll re-write and re-submit this PR later...

PR #5, opened 2023-05-22
ssh related fixes

hikalium:


src/dut.rs
```diff
@@ -698,7 +699,7 @@ impl SshInfo {
                     (child, port)
                 } else {
                     eprintln!("Failed to establish ssh port forwarding");
-                    return;
+                    exit(1);
```
mhiramat, line 702: Can't we return some error instead of forcibly exit? Or, at least it should be commented at the head of the function.

hikalium: Updated to return error when the first connection attempt is failed.

mhiramat (approved):
Looks good to me.

PR #9, opened 2023-05-31
Fix stall on reconnecting ssh port forwarding

hikalium:
BUG=None
TEST=reboot DUT while `lium chroot --dut ${DUT}`
TEST=and the connection was re-established.


hikalium:
I found that I broke the background ssh port forwarding in the previous commit (sorry...) so this is a fix for it. Please take a look!

mhiramat (approved):
Looks good to me.

PR #32, opened 2023-06-28
Make --cached option as a switch

mhiramat:
Commit 6635370 ("Add packages list subcommand") introduced --cached option as an option must be set. But it should be a switching option so that it can be disabled by default.


HidenoriKobayashi:
Ah, makes sense. LGTM.

Sorry that I didn't catch it in the previous PR. I wasn't familiar with argh and didn't bother to check it. But now I've read the document and I think `switch` is a better ... eh, option :smile:

hikalium (approved):

Issue #14, opened 2023-06-26
pass packages list as positional for `deploy` and `build` 


hikalium:
If we pass the packages list as positional argument, we can make it completable.

Let me explain more details.

The bash-completion is implemented by passing all available values to compgen and compgen filters the list using current argument as a prefix.

For example, if a command will get AAA BB BBC and CCC as an option, and if user inputs

command A[tab]
then, the completion script pass AAA BB BBC CCC to compgen. And compgen filters the list and returns only AAA.

If an option also takes AAA BB BBC CCC, same things happens.

But if the option takes a list of those, things become more complex. For example,

command --option "AAA BB[tab]
In this case, the completion script must pass a list of AAA AAA, AAA BB, AAA BBC and AAA CCC. And if

command --option "AAA BB [tab]
(there is a space separator after BB), then AAA BB AAA, AAA BB BB, AAA BB BBC and AAA BB CCC have to be passed. So it has to handle the space separators in the string in bash script. Also, if user add more options, the candidate list becomes longer.

But if we change don't use --option but just pass it as positional argument, we don't care about that.

command AAA BB [tab]
in any case, the completion script can pass AAA BB BBC CCC to compgen.

PR #33, opened 2023-06-28
14 pass packages list as positional for deploy and build

mhiramat:
This is for the issue #14 . Changing packages list as positional argument for  `lium build` and `lium deploy`, also update lium.bash to complemnt the package name.


HidenoriKobayashi (commented):
LGTM (I didn't look at lium.bash changes)

HidenoriKobayashi (approved):

PR #35, opened 2023-06-28
flash: Use the board name for looking up version

mhiramat:
Use the board name instead of "eve" for looking up version. 'lium sync' will continue use "eve".

### DIFF
--- a/src/cmd/flash.rs
+++ b/src/cmd/flash.rs
@@ -102,7 +102,7 @@ pub fn run(args: &Args) -> Result<()> {
     {
         args.version.clone()
     } else {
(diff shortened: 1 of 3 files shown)


HidenoriKobayashi:
LGTM.

Do we intend to get rid of all the use of "eve" eventually?

mhiramat:
For sync up to a specific version, I think we need some board name(s) to lookup version string as far as using current method.

hikalium:
maybe we should use manifest-versions in the future, but it looks like how the manifests are managed will be updated in the near future, so let's keep the current method for now ;)

LGTM!

hikalium (approved):

PR #37, opened 2023-06-28
tast: Add --option for test-specific options

mhiramat:
Add --option to `lium tast run` to specify test-specific options. (e.g. --option "-var")

### DIFF
--- a/src/cmd/tast.rs
+++ b/src/cmd/tast.rs
@@ -162,6 +162,10 @@ pub struct ArgsRun {
     #[argh(option)]
     dut: String,
 
+    /// test options (e.g. "-var ...")
+    #[argh(option)]
+    option: Option<String>,
+
     /// test name or pattern
     #[argh(positional)]
     tests: String,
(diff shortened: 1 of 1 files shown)


src/cmd/tast.rs
```diff
@@ -178,10 +182,19 @@ fn bundle_has_test(bundle: &str, filter: &Pattern) -> bool {
...
+    chroot: &Chroot,
+    port: u16,
+    opt: Option<&str>,
+) -> Result<()> {
     chroot.run_bash_script_in_chroot(
         "tast_run_cmd",
-        &format!("tast run -installbuilddeps -buildbundle={bundle} 127.0.0.1:{port} {filter}"),
+        &format!(
+            "tast run -installbuilddeps -buildbundle={bundle} {} 127.0.0.1:{port} {filter}",
+            opt.unwrap_or("")
```
HidenoriKobayashi, line 196: Hmm, the following is just a comment and I understand that this is a matter of preference. 

I think if we do unwrapping in the caller and just pass  `&str` to this function, it will be more compliant to the original function design.

HidenoriKobayashi: Yes, I think I understand your intention here and I'm not against it. I think it makes sense to pass an `Option` to `run_test_with_bundle`.

My take on this was that the original code seemed to process all the arguments in `run_tast_run`, so calling `unwrap_or` in `run_tast_run` and passing `&str` seemed closer to the original design. I did not mean that this is more preferred in Rust in general. Sorry for the confusion.

After all, this is just a comment and overall the change looks good to me :) Please feel free to resolve this conversation.

mhiramat: OK, thanks!

HidenoriKobayashi (approved):
LGTM

PR #44, opened 2023-07-06
lium.bash: Complete DUT name for `lium dut list --remove`

mhiramat:
Complete the (known) DUT name for `lium dut list --remove`

TEST=source lium.bash and `lium dut list --remove` [TAB]


hikalium (approved):
LGTM. Thanks!!

PR #46, opened 2023-07-06
lium deploy: add ab_update option

mhiramat:
Add --ab-update option to the lium deploy command. This option is valid only if the host kernel package is deployed.

### DIFF
--- a/src/cmd/deploy.rs
+++ b/src/cmd/deploy.rs
@@ -32,6 +32,10 @@ pub struct Args {
     /// if specified, it will skip automatic reboot
     #[argh(switch)]
     skip_reboot: bool,
+
+    /// use ab_update for kernel package
+    #[argh(switch)]
+    ab_update: bool,
 }
 pub fn run(args: &Args) -> Result<()> {
     ensure_testing_rsa_is_there()?;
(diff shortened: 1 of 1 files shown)


src/cmd/deploy.rs
```diff
@@ -85,8 +90,9 @@ pub fn run(args: &Args) -> Result<()> {
             &format!(
                 r###"
 cros-workon-{board} start {packages}
-~/trunk/src/scripts/update_kernel.sh --remote={} --ssh_port {} --remote_bootargs
+~/trunk/src/scripts/update_kernel.sh {} --remote={} --ssh_port {} --remote_bootargs
 "###,
+                ab_update,
```
hikalium, line 95: In Rust, if is an expression (= it can be a value) so you don't have to use a temporary variable here.
```
if args.ab_update { "--ab_update" } else { "" }
```
c.f. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0f36cb209de6f27c6d2aecab01ec67ff

hikalium (changes requested):

hikalium:
Except for the comment above, LGTM!

src/cmd/deploy.rs
mhiramat: Got it.

hikalium (approved):

PR #47, opened 2023-07-06
lium dut: Add MAC addr based serial number if no serial number

mhiramat:
Generate a MAC address based serial number and use it if the DUT doesn't have its serial number.

### DIFF
--- a/src/dut.rs
+++ b/src/dut.rs
@@ -179,7 +179,7 @@ const CMD_GET_DEFAULT_IFACE: &str =
     r"ip route get 8.8.8.8 | sed -E 's/^.* dev ([^ ]+) .*$/\1/' | head -n 1";
 
 // Only keys that are always available can be listed here
-const DEFAULT_DUT_INFO_KEYS: [&str; 7] = [
+const DEFAULT_DUT_INFO_KEYS: [&str; 8] = [
     "timestamp",
     "dut_id",
     "hwid",
@@ -187,6 +187,7 @@ const DEFAULT_DUT_INFO_KEYS: [&str; 7] = [
     "model",
     "serial",
     "board",
+    "mac",
 ];
 
 /// DutInfo holds information around a DUT
@@ -307,13 +308,26 @@ impl DutInfo {
             }
         }
         if keys.contains(&"dut_id") {
-            let model = values.get("model");
-            let serial = values.get("serial");
-            if let (Some(Ok(model)), Some(Ok(serial))) = (model, serial) {
+            let serial = if let Some(Ok(serial)) = values.get("serial") {
+                serial.to_string()
+            } else {
+                // Some DUTs don't have serial number. So use MAC address,
+                let serial = if let Some(Ok(mac)) = values.get("mac") {
+                    format!("NoSerial{}", mac.replace(':', "").to_lowercase())
+                } else {
+                    return Err(anyhow!(
+                        "Failed to get MAC address. {:?}",
+                        values.get("mac")
+                    ));
+                };
+                values.insert("serial".to_string(), Ok(serial.to_string()));
+                serial
+            };
+            if let Some(Ok(model)) = values.get("model") {
                 let dut_id = format!("{model}_{serial}");
                 values.insert("dut_id".to_string(), Ok(dut_id));
             } else {
-                return Err(anyhow!("model and serial is needed for dut_id but got: (model: {model:?}, serial: {serial:?})"));
+                return Err(anyhow!("Failed to get model. {:?}", values.get("model")));
             }
         }
         // Collect all values for given keys


src/dut.rs
```diff
@@ -307,13 +308,23 @@ impl DutInfo {
...
-            let serial = values.get("serial");
-            if let (Some(Ok(model)), Some(Ok(serial))) = (model, serial) {
+            let serial: String = if let Some(Ok(serial)) = values.get("serial") {
+                serial.to_string()
+            } else {
+                // Some DUTs don't have serial number. So use MAC address,
+                let serial = if let Some(Ok(mac)) = values.get("mac") {
+                    mac.replace(':', "")
+                } else {
+                    return Err(anyhow!("Failed to get MACaddr. {:?}", values.get("mac")));
```
hikalium, line 318: nit: `MAC address` may be better

src/dut.rs
```diff
@@ -307,13 +308,23 @@ impl DutInfo {
...
         if keys.contains(&"dut_id") {
-            let model = values.get("model");
-            let serial = values.get("serial");
-            if let (Some(Ok(model)), Some(Ok(serial))) = (model, serial) {
+            let serial: String = if let Some(Ok(serial)) = values.get("serial") {
+                serial.to_string()
+            } else {
+                // Some DUTs don't have serial number. So use MAC address,
+                let serial = if let Some(Ok(mac)) = values.get("mac") {
+                    mac.replace(':', "")
```
hikalium, line 316: how about adding some prefix to explicitly state "this is not a real serial number" to avoid confusion?
e.g. "NoSerial" so the "serial" will be "NoSerialAABBCCDDEEFF"

hikalium: Also, macaddr should be canonicalized to lowercase to avoid issues later (e.g. matched DUT not found since both upper / lower cases are used in different datasources)
c.f. https://www.ieee802.org/1/files/public/docs2020/yangsters-smansfield-mac-address-format-0420-v01.pdf

src/dut.rs
```diff
@@ -307,13 +308,23 @@ impl DutInfo {
...
+                } else {
+                    return Err(anyhow!("Failed to get MACaddr. {:?}", values.get("mac")));
+                };
+                values.insert("serial".to_string(), Ok(serial.to_string()));
```
hikalium, line 320: the "pseudo serial" should not be stored as serial here, as it is not a serial. (we should use dut_id everywhere to identify a DUT)

hikalium (changes requested):

hikalium:
Generally speaking, this is exactly what I was thinking to implement. Thanks for making the PR!

src/dut.rs
mhiramat: Agreed.

src/dut.rs
mhiramat: Hmm, what about "MAC_AABBCCDDEEFF" for short?
Also, it already uses lowercase for "mac" value because `ip addr` shows it in lowercase.

src/dut.rs
mhiramat: But without serial, the command doesn't work. See https://github.com/google/lium/blob/4f24e79a498d17a4de9ed762700eafa2f17dcec2/src/dut.rs#L326

src/dut.rs
hikalium: I want to avoid using "_" in the "serial" part of dut_id since sometimes model name includes "_" (e.g. sarien_signed) and it makes it difficult to separate the model part and serial part.

Also, there is no validation that mac is lowercase, so making sure they are lower case here is a good practice. (ideally, the length of the mac addr should be checked as well...)

And, it looks like there is a serial number that contains a string "mac" like "0MAC91ZH..." so just using MAC as a prefix may be confusing. I understand shorter is better, but with the command completion, it should be not a huge problem.

What do you think...?

mhiramat: At first, I think canonicalizing the mac address is another issue, and should be done where the "mac" value is got.

And I wonder if we should be so cautious about using (temporary) mac addresses as serial numbers. What do you think a problematic scenario?

hikalium: > At first, I think canonicalizing the mac address is another issue, and should be done where the "mac" value is got.
yes, we should do the same thing on the logic that actually getting mac addr from the DUT. However, doing the same thing again here is not harmful, so I recommend doing that here. 

> And I wonder if we should be so cautious about using (temporary) mac addresses as serial numbers. What do you think a problematic scenario?

First, we want to make sure the "temporary" DUT_ID does not collide with the normal, valid "DUT_ID", since it can cause an inaccessible DUT when using a DUT_ID for `lium dut shell` or something. This is why we want to have a distinguishable prefix for such a pseudo-serial number.  Second, without output validation and if for some reason the command returns mac addr based DUT_ID in uppercase / lowercase randomly, you can't ssh sometimes because of the DUT_ID mismatch. For example, 
```
lium dut shell --dut `lium dut info --dut ${IP} | jq -r .dut_id` -- echo "OK"
```
should always succeed for all the valid `${IP}`, but if the DUT_ID returned by `lium dut info` is not stable, it cannot be accomplished.

src/dut.rs
hikalium: That is because serial will be the dependency of dut_id (which is not true anymore with this change)

https://github.com/google/lium/blob/ca94eaad10db333982e2afe94bd73e64b84fa1b4/src/dut.rs#L345

so you can remove the line above to make it work ;)

src/dut.rs
mhiramat: Ok, so you might concern that the existing serial number is similar (or same as) the mac address. Let me adding `NoSerial`.

src/dut.rs
mhiramat: If I removed only that requirement, `lium dut list --add` failed. If `serial` is not found.
```
Error: failed to get key serial: Some(Err(key serial found but was empty. stderr: [WARN] VPD partition not formatted. It's fine.
findString(): Vpd data 'serial_number' was not found.))
```
As I pointed, the `parse_values()` checks the `DEFAULT_DUT_INFO_KEYS` at the last loop, and if it returns an error, it shows the error.

src/dut.rs
```diff
@@ -307,13 +308,24 @@ impl DutInfo {
             }
         }
         if keys.contains(&"dut_id") {
-            let model = values.get("model");
-            let serial = values.get("serial");
-            if let (Some(Ok(model)), Some(Ok(serial))) = (model, serial) {
+            let serial: String = if let Some(Ok(serial)) = values.get("serial") {
```
HidenoriKobayashi, line 311: nit: do we need an explicit type (`String`) here?

Seems like it's obvious for the compiler because of the `to_string()`.

src/dut.rs
hikalium: https://github.com/google/lium/blob/ca94eaad10db333982e2afe94bd73e64b84fa1b4/src/dut.rs#L181-L190

Ah, then please remove `serial` from `DEFAULT_DUT_INFO_KEYS` as well, since the assumption (`serial` is one of the "keys that are always available") was wrong.

src/dut.rs
mhiramat: Thanks, removed.

src/dut.rs
mhiramat: No, it doesn't work.
```
$ lium dut info --dut geralt_NoSeriala0cec8e67395 
Fetching info for SshInfo { host: "2401:fa00:480:ee08:a2ce:c8ff:fee6:7395", port: 22 }...
Error: failed to get key serial: Some(Err(key serial found but was empty. stderr: [WARN] VPD partition not formatted. It's fine.
findString(): Vpd data 'serial_number' was not found.))
```
With following change.
```
--- a/src/dut.rs
+++ b/src/dut.rs
@@ -179,13 +179,12 @@ const CMD_GET_DEFAULT_IFACE: &str =
     r"ip route get 8.8.8.8 | sed -E 's/^.* dev ([^ ]+) .*$/\1/' | head -n 1";
 
 // Only keys that are always available can be listed here
-const DEFAULT_DUT_INFO_KEYS: [&str; 8] = [
+const DEFAULT_DUT_INFO_KEYS: [&str; 7] = [
     "timestamp",
     "dut_id",
     "hwid",
     "release",
     "model",
-    "serial",
     "board",
     "mac",
 ];
@@ -320,7 +319,6 @@ impl DutInfo {
                         values.get("mac")
                     ));
                 };
-                values.insert("serial".to_string(), Ok(serial.to_string()));
                 serial
             };
             if let Some(Ok(model)) = values.get("model") {
```

hikalium: ok, let 's do so...

hikalium (approved):

Issue #40, opened 2023-07-05
Pipe-processing friendly output for servo list output


cynthia:
Right now, the default output for `servo list` is JSON.

It would be nice if we didn't need to rely on arcane methods to massage the output into something usable. For example, here is how you get a list of the usb devices so that it can be processed by a shell script:

    lium servo list | jq -r '.[] | map(.serial + ": " + .usb_sysfs_path)[]' \
    | awk '{print $2 " " $1;}' | sort | tr -d ':'

Given that this is a CLI first, it'd make sense to make the default output pipe-processing friendly, and make JSON output an opt-in (as that would be more likely something that would be machine-processed).

hikalium:
Thanks for the issue! Actually, I just started working on updating the interfaces of servo related commands as Unix-commands-friendly and machine-friendly with --json flag. Let me take the issue and work on a fix.

PR #53, opened 2023-07-12
Servo related updates

hikalium:
This fixes:
- https://github.com/google/lium/issues/40
- https://github.com/google/lium/issues/52


src/cmd/setup.rs
```diff
@@ -159,44 +160,45 @@ fn setup_dut(repo: &str, cr50: &LocalServo) -> Result<()> {
...
-            continue;
-        }
+    let servo = list.find_by_serial(&args.serial).context(
+        "No Servos or Cr50 are detected. Please check the servo connection, try another side of USB port, attach servo directly with a host instead of via hub, etc...")?;
+    eprintln!("Using {servo:?}");
+    if args.ccd_unlock {
+        let cr50 = get_cr50_attached_to_servo(servo)?;
+        setup_dut_ccd_open(&cr50)
+    } else {
+    setup_dut(&repo, servo)
```
mhiramat: nit: miss-indent

src/cmd/setup.rs
```diff
@@ -15,7 +15,7 @@ use lium::cros::ensure_testing_rsa_is_there;
 use lium::dut::register_dut;
 use lium::dut::DutInfo;
 use lium::repo::get_repo_dir;
-use lium::servo::get_servo_attached_to_cr50;
+use lium::servo::get_cr50_attached_to_servo;
```
mhiramat, line 18: Which did change this function name?

src/cmd/setup.rs
hikalium: true, but it is fixed by the later commit. please review this PR as a whole... (I think GitHub's pull-request system is designed in that way, as you can see the diff above is marked as "Outdated") (but thanks for pointing it out!)

src/cmd/setup.rs
```diff
@@ -149,7 +158,8 @@ fn ensure_dev_gbb_flags(repo: &str, cr50: &LocalServo) -> Result<()> {
     Ok(())
 }
 
-fn setup_dut(repo: &str, cr50: &LocalServo) -> Result<()> {
+fn setup_dut(repo: &str, servo: &LocalServo) -> Result<()> {
```
hikalium, line 161: because of this function changed to take servo instead of cr50...

src/cmd/setup.rs
```diff
@@ -149,7 +158,8 @@ fn ensure_dev_gbb_flags(repo: &str, cr50: &LocalServo) -> Result<()> {
...
+    let cr50 = get_cr50_attached_to_servo(servo)?;
```
hikalium, line 162: conversion from cr50 to servo is needed

src/cmd/setup.rs
```diff
@@ -159,44 +169,44 @@ fn setup_dut(repo: &str, cr50: &LocalServo) -> Result<()> {
     }
 
     eprintln!("Setting up DUT: {}", cr50.serial());
-    setup_dut_ccd_open(cr50)?;
-    let mut servo = get_servo_attached_to_cr50(cr50)?;
```
hikalium, line 163: and the other way is removed

src/cmd/servo.rs
```diff
@@ -116,19 +115,24 @@ pub fn run_reset(args: &ArgsReset) -> Result<()> {
 /// list servo-compatible devices (Servo V4, Servo V4p1, SuzyQable)
 #[argh(subcommand, name = "list")]
 pub struct ArgsList {
-    /// update the cached servo info. It will take a few seconds per servo.
+    /// print additional info as well (takes more time)
     #[argh(switch)]
-    update: bool,
+    extra: bool,
```
HidenoriKobayashi, line 120: What does this `extra` actually print out in addition? I tried but I'm not getting any additional information.
Is this for future extension?

src/cmd/servo.rs
```diff
@@ -138,7 +142,16 @@ pub fn run_list(args: &ArgsList) -> Result<()> {
...
         return Ok(());
     }
-    println!("{}", list);
+    if args.json {
+        println!("{}", list);
+        return Ok(());
+    }
+    println!("usb_sysfs_path\tserial\tproduct");
+    let mut devices = list.devices().clone();
+    devices.sort_by(|l, r| l.usb_sysfs_path().cmp(r.usb_sysfs_path()));
```
HidenoriKobayashi, line 151: I guess we only sort the result for human friendly format but not for json.
If it doesn't hurt to have them sorted in both cases, it might be better to sort them when you create the list in ServeList. Although, I'm not sure even for human readable cases, what order is the best way to present this kind of information. (This is just a comment.)

src/cmd/servo.rs
```diff
@@ -138,7 +142,16 @@ pub fn run_list(args: &ArgsList) -> Result<()> {
...
+    for s in devices {
+        println!("{}\t{}\t{}", s.usb_sysfs_path(), s.serial(), s.product());
```
HidenoriKobayashi, line 153: I probably want to it in the following order: product, serial, sysfs_path.

But this is a highly subjective matter, I guess. If the current order makes sense for you, that's fine.

By the way, it would be nice to have the column aligned, meaning determining a reasonable width for each field instead of just using `\t`. Right now, it looks like this for me:
```
usb_sysfs_path	serial	product
/sys/bus/usb/devices/1-6.1	G1911051939	Servo V4
/sys/bus/usb/devices/1-6.2	0780605B-8D25B743	Cr50
```

src/cmd/setup.rs
mhiramat: I doubt the github PR works such way. I think PR is a series of patches because I can not review a whole change. Anyway, it is just a nitpick.

src/cmd/setup.rs
mhiramat: Ah, OK. There are 2 similar functions and just switched it.

src/cmd/setup.rs
mhiramat: OK, I got it.

src/cmd/setup.rs
mhiramat: OK, I got it.

src/cmd/servo.rs
mhiramat: It seems `args.extra` only changes `ServoList::discover_slow()` or `ServoList::discover()` and that will not change the printing result. I think old `--update` explains better.

src/cmd/servo.rs
hikalium: Thanks for the suggestion! Now it looks like this as suggested:
```
(workstation) $ lium servo list
product         serial                          usb_sysfs_path
Servo V4p1      SERVOV4P1-S-2202180988          /sys/bus/usb/devices/1-4.5
```

src/cmd/servo.rs
hikalium: Made the order stable for all the output options.

src/cmd/servo.rs
hikalium: Since now ServoList is not cached, --update is a bit misleading so I changed it to extra, since it will retrieve extra info.
And the change will be like this:

without extra
```
(workstation) $ lium servo list --json
{
  "devices": [
    {
      "product": "Servo V4p1",
      "serial": "SERVOV4P1-S-2202180988",
      "usb_sysfs_path": "/sys/bus/usb/devices/1-4.5",
      "tty_list": {
        "Atmega UART": "/dev/ttyUSB3",
        "DUT UART": "/dev/ttyUSB2",
        "Firmware update": "/dev/ttyUSB4",
        "I2C": "/dev/ttyUSB1",
        "Servo EC Shell": "/dev/ttyUSB0"
      },
      "cached_info": null
    }
  ]
}
```

with extra
```
(workstation) $ lium servo list --json --extra
Checking SERVOV4P1-S-2202180988
{
  "devices": [
    {
      "product": "Servo V4p1",
      "serial": "SERVOV4P1-S-2202180988",
      "usb_sysfs_path": "/sys/bus/usb/devices/1-4.5",
      "tty_list": {
        "Atmega UART": "/dev/ttyUSB3",
        "DUT UART": "/dev/ttyUSB2",
        "Firmware update": "/dev/ttyUSB4",
        "I2C": "/dev/ttyUSB1",
        "Servo EC Shell": "/dev/ttyUSB0"
      },
      "cached_info": {
        "mac_addr": "88:54:1f:0f:72:fe"
      }
    }
  ]
}
```

cached_info will be cached in the future, but not at this point... (I'll think about the caching feature later)
... okay, let me change it back to --update (to align with the output...)

hikalium: now they are "SlowInfo" and the flag is "--slow"

HidenoriKobayashi (commented):
CR+1.

The change looks good to me now, but I think I want wait for @mhiramat to complete his review.

src/cmd/servo.rs
```diff
@@ -138,7 +142,20 @@ pub fn run_list(args: &ArgsList) -> Result<()> {
         println!("{}", keys.join(" "));
         return Ok(());
     }
-    println!("{}", list);
+    if args.json {
+        println!("{}", list);
+        return Ok(());
+    }
+    println!("product         serial                          usb_sysfs_path");
```
mhiramat, line 149: isn't this use same format (offset values) as below code?
```
println!("{:16}{:24}\t{}", "product", "serial", "usb_sysfs_path");
```

hikalium: Clippy complains if I use the format string with static strings... that's why I chose this way...

HidenoriKobayashi: true, although it might become more readable/maintainable with what clippy suggests:
```
println!("{:16}{:24}\tusb_sysfs_path", "product", "serial");
```

mhiramat: Hmm, I think clippy missed the point, if we change some items to display, or change the order, that format can not simply swap it. Using the same format string is the key point, so that we can change the data easily.

mhiramat: OK, anyway it seems hard to escape from the clippy optimization. So after introducing some better formatting (for showing table) we will revisit it.

mhiramat (approved):

PR #54, opened 2023-07-15
Add servo_power_meter

hikalium:
Import a power measurment tool from github.com/hikalium/hikalium.com.


cynthia:
The dark scheme is sort of an ugly hack. I don't mind if you make it part of mainline - but I'll have to send a couple patches to fix some known issues.

cynthia (approved):
Committing build artifacts is a bit of a process-smell, but it's probably fine to clean up later with actions.

PR #61, opened 2023-07-28
dut: add reboot_other

HidenoriKobayashi:
This adds a new sub-sub-command to reboot into the other partition.

This shouldn't be terribly slow for local DUTs.

### DIFF
--- a/src/cmd/dut.rs
+++ b/src/cmd/dut.rs
@@ -241,6 +241,45 @@ type DutAction = Box<fn(&SshInfo) -> Result<()>>;
 fn do_reboot(s: &SshInfo) -> Result<()> {
     s.run_cmd_piped(&["reboot; exit"])
 }
+
+enum PartitionSet {
+    Primary,
+    Secondary,
+}
+fn switch_partition_set(s: &SshInfo, target: PartitionSet) -> Result<()> {
+    let rootdev = s.get_rootdev()?;
+    let rootdisk = s.get_rootdisk()?;
+    let part = s.get_partnum_info()?;
+    let kern_a = part.get("kern_a").ok_or(anyhow!("KERN-A not found"))?;
+    let kern_b = part.get("kern_b").ok_or(anyhow!("KERN-B not found"))?;
+    let root_a = part.get("root_a").ok_or(anyhow!("ROOT-A not found"))?;
+    let root_b = part.get("root_b").ok_or(anyhow!("ROOT-B not found"))?;
+    let (current_name, current_kern, current_root, other_name, other_kern, other_root) =
+        if rootdev.ends_with(root_a) {
+            ("A", kern_a, root_a, "B", kern_b, root_b)
+        } else if rootdev.ends_with(root_b) {
+            ("B", kern_b, root_b, "A", kern_a, root_a)
+        } else {
+            return Err(anyhow!("unsupported partition layout"));
+        };
+    let cmd = match target {
+        PartitionSet::Primary => {
+            println!("switching to primary: {current_name} ({rootdisk}p{current_root})");
+            format!("cgpt prioritize -P2 -i {current_kern} {rootdisk}")
+        }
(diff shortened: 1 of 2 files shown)


src/cmd/dut.rs
```diff
@@ -241,6 +241,26 @@ type DutAction = Box<fn(&SshInfo) -> Result<()>>;
...
+    } else if rootdev.ends_with(&root_b) {
+        ("A", &kern_a, &root_a)
+    } else {
+        return Err(anyhow!("unsupported partition layout"));
+    };
+    println!("rebooting into {other_name} ({rootdisk}p{other_root})");
+    let cmd = format!("cgpt prioritize -P2 -i {other_kern} {rootdisk}");
+    s.run_cmd_stdio(&cmd)?;
+
+    do_reboot(s)
```
hikalium, line 262: I think it is a good idea to separate this command into two:
```
toggle_boot_partition reboot
```

src/cmd/dut.rs
```diff
@@ -260,6 +280,7 @@ lazy_static! {
         let mut m: HashMap<&'static str, DutAction> = HashMap::new();
         m.insert("wait_online", Box::new(do_wait_online));
         m.insert("reboot", Box::new(do_reboot));
+        m.insert("reboot_other", Box::new(do_reboot_other));
```
hikalium, line 283: so replace this with `toggle_boot_partition`

src/dut.rs
```diff
@@ -747,6 +747,32 @@ impl SshInfo {
...
+        self.run_cmd_stdio(cmd_str)
+    }
+    pub fn get_partnum_kern_a(&self) -> Result<String> {
+        let cmd_str = "source /usr/sbin/write_gpt.sh; load_base_vars; echo ${PARTITION_NUM_KERN_A}";
+        self.run_cmd_stdio(cmd_str)
+    }
+    pub fn get_partnum_kern_b(&self) -> Result<String> {
+        let cmd_str = "source /usr/sbin/write_gpt.sh; load_base_vars; echo ${PARTITION_NUM_KERN_B}";
+        self.run_cmd_stdio(cmd_str)
+    }
```
hikalium, line 775: For `get_partnum_*`, How about just getting load_base_vars variables at once, not by each items...?

hikalium (changes requested):

src/cmd/dut.rs
HidenoriKobayashi: removed the reboot part

src/dut.rs
HidenoriKobayashi: combined them

src/cmd/dut.rs
HidenoriKobayashi: I changed it to "switch_to_secondary", since what the command does is not actually toggling the partition. If you run the command twice, the higher priority is still given to the secondary partition.

hikalium: I see. In that case, how about adding a switch for the opposite direction (boot from primary) when the user wants to get back to the original one? (I'm not sure how you can get back to the normal boot order without adding such command (except for manual re-flashing...))

HidenoriKobayashi: Okay, I can add another sub-command for that. I think it is better than actually implementing the toggle command, which is doable but maybe more confusing since that forces you to know the current priority order.  In contrast, `switch_to_primary/secondary` sub-commands can be issued without knowing that.
If you really want to go for the `toggle`, please let me know.

---
Well, in theory, you can reboot the DUT and `switch_to_secondary` and reboot again to get back to the original set... (Yeah, of course, I do not mean it)

HidenoriKobayashi: I just uploaded a version with `switch_to_primary` and `switch_to_secondary`.  Does this look better?

hikalium: yes, it looks good to me now. thanks for the fix!

PR #62, opened 2023-08-02
flash: Support --image arg to specify local chromiumos image file

hikalium:
TEST=lium flash --usb --image ${PATH_TO_IMAGE}

### DIFF
--- a/src/cmd/flash.rs
+++ b/src/cmd/flash.rs
@@ -35,6 +35,10 @@ pub struct Args {
     #[argh(option)]
     board: Option<String>,
 
+    /// path to image to flash
+    #[argh(option)]
+    image: Option<String>,
+
     /// chromiumos version to flash (default: latest-dev)
     #[argh(option, default = "String::from(\"latest-dev\")")]
     version: String,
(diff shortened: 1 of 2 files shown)


src/cmd/flash.rs
```diff
@@ -55,62 +59,67 @@ pub fn run(args: &Args) -> Result<()> {
...
-                // The board name may have suffix '64' or '-*', so get the first alphabet
-                // sequence as the base board name.
-                let re = Regex::new(r"(^[[:alpha:]]*)")?;
-                let cap_arg = if let Some(cap) = re.captures(board_from_arg) {
-                    cap
+    let image_path = if let Some(image) = &args.image {
+        // If --image is specified, use the local file
+        image.clone()
+    } else {
+        // Determine a BOARD to flash
```
mhiramat, line 66: Can you make this `else` block a local function?

hikalium: The part is extracted into determine_board_to_flash(). Please review again!

HidenoriKobayashi: I'm curious if the original comment meant  the whole thing in the else block or the part that was extracted as `determine_board_to_flash`

hikalium: Not sure what do you mean... @mhiramat Could you checkif the change is acceptable and aligned with your suggestion? Thanks!!!

hikalium: ping...

hikalium:
@mhiramat could you review this again? Thank you!

src/cmd/flash.rs
mhiramat: Sorry for later. Yes, @HidenoriKobayashi is right. The else block seems independently works.

mhiramat (approved):
This looks good to me. Thanks!

PR #63, opened 2023-08-02
dut: setup: move lium setup dut command and improve stability

hikalium:


HidenoriKobayashi:
Could you kindly provide us a bit of information to help reviewing this?

I understand that in theory we should be reading every changed line, but if you could let us know which function has been newly introduced (or which functions were just moved) we could probably give you more valuable feedback.

hikalium:
Basically, all the code in src/cmd/setup.rs that is related to `lium setup dut` command before this commit is moved into src/cmd/dut.rs. Also, changed servo reset order in a few places to make it stable and fast. Plus, added flags like `--get_ccd_status` to run each step manually, for debugging and recovering after the failures.

hikalium:
I think the change is ready to be merged. Please take a look again!

src/cmd/dut.rs
```diff
@@ -265,6 +272,287 @@ lazy_static! {
...
+        reset_gbb_flags(&repo, &cr50)?;
+    } else if args.enable_ccd_testlab {
+        enable_ccd_testlab(&cr50)?;
+    } else if args.check_ssh {
+        check_ssh(&servo)?;
+    } else {
+        open_ccd(&cr50)?;
+        enable_ccd_testlab(&cr50)?;
+        let dut = check_ssh(&servo)?;
+        check_gbb_flags(&dut)?;
```
HidenoriKobayashi, line 551: Is there any reason why this function cannot be run directly with an option? All the other functions called from this else block seem to have an option to run it.

hikalium: Good point! Added a new commit that calls the check_gbb_flags function only. Please review again! (sorry for the delayed response...)

src/cmd/dut.rs
```diff
@@ -265,6 +272,296 @@ lazy_static! {
...
+/// Check if GBB flags are set for development.
+fn check_dev_gbb_flags(dut: &DutInfo) -> Result<()> {
+    let info = DutInfo::fetch_keys(dut.ssh(), &vec!["gbb_flags"])?;
+    let gbb_flags = info
+        .get("gbb_flags")
+        .context("gbb_flags is not set")?
+        .replace(r"0x", "");
+    let gbb_flags = u64::from_str_radix(&gbb_flags, 16).context("failed to parse gbb_flags")?;
+    eprintln!("GBB flags: {gbb_flags:#10X}");
+    // 
```
HidenoriKobayashi, line 461: Is this due to GitHub treating some Japanese characters wildly?

HidenoriKobayashi (approved):
I want to confirm about l. 461 in dut.rs, but otherwise LGTM.

PR #64, opened 2023-08-17
Release v0.1.1

hikalium:
Created a release on crates.io:
https://crates.io/crates/lium


hikalium:
With this change, `make release` will create a binary release on GitHub, like https://github.com/google/lium/releases/tag/v0.1.1 . Please take a look!

scripts/deploy_release.sh
```diff
@@ -0,0 +1,32 @@
...
+cargo build --release --target-dir ${TMPDIR}
+VERSION=`${TMPDIR}/release/lium version | cut -d ' ' -f 2`
+
+if gh release list | grep ${VERSION} ; then
+	# Release with the version found. Delete it first.
+	gh release delete -y ${VERSION}
+fi
+
+# Create a new release
+read -r -d '' RELEASE_NOTE <<EOF
```
HidenoriKobayashi, line 26: I see that a user can figure out the version from the curl command line, but still it might be a good idea to add another line stating the version explicitly in the first line. e.g. LIUM version v0.1.1.

Although I guess it's not strictly necessary at this point. Feel free to close this comment by yourself.

hikalium: I think we don't need that for now and it can be easily noticed by seeing the command line. it would be appreciated if you allow me to proceed with this.

HidenoriKobayashi (approved):
LGTM :)

PR #73, opened 2023-09-29
enable WebSerial start button twice

wtrgjadmw:
**What I have done:**

- make `requestSerialButton` disabled while measuring WebSerial power
- make serial port closed after stopping measurement


docs/servo_power_monitor/src/index.ts
```diff
@@ -165,47 +165,47 @@ function setupStartUSBButton() {
...
-          } finally {
-            reader.releaseLock();
-          }
-        });
-      })
-      .catch((e) => {
-        // The user didn't select a port.
-        console.error(e);
-      });
+  requestSerialButton.disabled = true;
```
hikalium, line 173: I think it's too early to disable the button, since it's possible that the user cancels the port selection window and the user can't select a port anymore since the button is disabled. how about moving the line after the port selection?

Also, It's a good idea to think what will happen if you disconnect the servo during the measurement is ongoing... I think the button will be kept disabled and the user have to reload the page to restart it again.

Thinking about error case is an important thing ;) (but yeah, it's difficult, so please try!)

hikalium (changes requested):
One nit, otherwise looks good to me!

wtrgjadmw:
I checked it works in the following situation;
- start(serial) -> stop, disconnect: close serial port 
- start(usb) -> stop, disconnect: close usb port

After both situations, power can be measured again.

docs/servo_power_monitor/src/index.ts
```diff
@@ -94,22 +94,47 @@ function kickWriteLoop(writeFn: (s: string) => Promise<void>) {
...
-    if (s === undefined || !s.length) {
-      continue;
+    try {
+      const s = await readFn();
+      if (s === undefined || !s.length) {
+        continue;
+      }
+      pushOutput(s);
+    } catch (e) {
+      break;
```
hikalium, line 104: Adding a comment to explain why we can just ignore the error here will be useful for future developers ;)

docs/servo_power_monitor/src/index.ts
```diff
@@ -122,26 +147,33 @@ function setupStartUSBButton() {
     }
     if (!device) {
       device = null;
-      requestUSBButton.disabled = false;
       return;
     }
-
+
```
hikalium, line 153: please delete this line (it seems to be irrerevant)

hikalium: clang-format can help you styling the code and fix issues like this automatically, please try it out ;)

docs/servo_power_monitor/src/index.ts
```diff
@@ -122,26 +147,33 @@ function setupStartUSBButton() {
...
+        try {
+          let result = await device.transferIn(ep, 64);
+          if (result.status === 'stall') {
+            await device.clearHalt('in', ep);
+            throw result;
+          }
+          const result_array = new Int8Array(result.data.buffer);
+          return utf8decoder.decode(result_array);
+        } catch (e) {
+          if (!halt) {
```
hikalium, line 173: Adding a comment to explain why we are checking the halt flag here and why it is safe to ignore the error when halt is true is a good practice

docs/servo_power_monitor/src/index.ts
```diff
@@ -166,47 +198,63 @@ function setupStartUSBButton() {
...
+    } catch (error) {
+      console.error(error);
+      throw error;
+    } finally {
+      reader.releaseLock();
+    }
+  });
+});
+
+// event when you disconnect USB port
```
hikalium, line 243: nit: I would say:
```
// `disconnect` event is fired when a USB device is disconnected.
// c.f. https://wicg.github.io/webusb/#disconnect
```

docs/servo_power_monitor/src/index.ts
wtrgjadmw: Thank you for your advice, I added comments!

docs/servo_power_monitor/src/index.ts
```diff
@@ -166,47 +198,63 @@ function setupStartUSBButton() {
...
+navigator.usb.addEventListener("disconnect", () => {
+  if (requestUSBButton.disabled) {
+    halt = true;
+    requestUSBButton.disabled = false;
+    inProgress = false;
+    // USB port is closed by specification when device is disconnected
```
hikalium, line 249: nit: I would say:
```
//  No need to call close() for the USB port here because the specification says that
// the port will be closed automatically when a device is disconnected.
// c.f. <paste the URL to a section in the spec which says so>
```

wtrgjadmw: maybe the nearest URL is https://wicg.github.io/webusb/#disconnect, so I removed the last sentence.

wtrgjadmw:
I fixed comments!

hikalium (commented):
Roughly looks good to me!

Only two comments that can improve your pull request furthermore!

- Generally, it's a good practice to separate the project-wide coding style fixes and functional changes to help reviewers and future developers to understand what is changed. This time, it is clear from the commit history that which change was the style update, so you can proceed with this. However, in the future, you should file a separate pull request for such style changes.

- when adding an inline comment to explain the code, adding the comment above the part which you are explaining is better than putting it below the code. This is for consistency with other engineers. Also, if you are writing some explanation for the part of the conditionally executed code, put the comment in the block of the same level as the conditionally executed part is a good idea.

For example,

```
// comment for something
Do something();
```

```
if condition {
  // comment for the true case
  DoSomethingWhenTrue();
} else {
  // comment for the false case
  DoSomethingWhenFalse();
}
```

wtrgjadmw:
Thank you hikalium, I just fixed the position of comments for now.
By the way, I have one question about your first advice. Which does the situation mean formatting a file (e.g. formatting index.ts) or change style for the project overall (e.g. adding [gts](https://github.com/google/gts))?

hikalium:
As I chatted offline, "the project-wide coding style fixes" and "functional changes" are usually separated logically, so they should be filed as separated pull-requests to make the review easier. Anyway, now your change looks good for me, so I merge it into the main branch. Thanks for the great work!!

hikalium (approved):

Issue #74, opened 2023-09-29
Create a rustfmt.toml and reformat


brandonpollack23:
This is based on my own rustfmt.toml I usually use.  I  tend to set these for the following reasons:

imports_granularity = "Crate"  # this combines a bunch of lines at the top, but it doesnt match google style so I removed it
tab_spaces = 2 # param lists and chained calls tend to be longer and more common in rust, I think this increases readability
fn_params_layout = "Tall" # this is way easier to read than non tall
format_strings = true
use_try_shorthand = true # replace any ugly try! macros with the pretty ?
wrap_comments = true 
max_width = 100 # If we can do it in Java, we can do it in Rust

brandonpollack23:
PR #76

PR #76, opened 2023-09-29
Add a rustfmt.toml file with some clean guidlines.

brandonpollack23:
Note: Though seemingly large this change is a total no-op and run with `cargo fmt` only, just he simplest thing I could think of for a first issue.

This is based on my own rustfmt.toml I usually use. I tend to set these for the following reasons:

imports_granularity = "Crate" # this combines a bunch of lines at the top, but it doesnt match google style so I removed it tab_spaces = 2 # param lists and chained calls tend to be longer and more common in rust, I think this increases readability fn_params_layout = "Tall" # this is way easier to read than non tall format_strings = true
use_try_shorthand = true # replace any ugly try! macros with the pretty ? wrap_comments = true
max_width = 100 # If we can do it in Java, we can do it in Rust

for issue #74

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,6 @@
+tab_spaces = 4
+fn_args_layout = "Tall"
+format_strings = true
+use_try_shorthand = true
+wrap_comments = true
+max_width = 100
--- a/src/chroot.rs
+++ b/src/chroot.rs
@@ -81,7 +81,8 @@ impl Chroot {
     ///
     /// # Arguments
     ///
-    /// * `name` - A filename-capable identifier for the script (this file will be created automatically)
+    /// * `name` - A filename-capable identifier for the script (this file will
+    ///   be created automatically)
     /// * `script` - One or more lines of bash script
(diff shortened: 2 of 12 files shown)


rustfmt.toml
```diff
@@ -0,0 +1,6 @@
+tab_spaces = 2
```
HidenoriKobayashi, line 1: Hmm, I prefer keeping the tab_spaces to the default 4.

HidenoriKobayashi (commented):
I'm mostly fine with this change, except for the `tab_spaces`.

Although, I can probably live with it, if everybody else prefers it to be 2.

rustfmt.toml
brandonpollack23: 4 is the default style so let's stick with that, (all this is just to learn to use github anyways).  Done!

brandonpollack23:
Closing this PR because it is included in the "Starter Changes PR"

Feel free to comment on that for the PR to be split up at a certain point.

PR #80, opened 2023-09-29
Setup git-ps to do stacked reviews on github.

brandonpollack23:
This will allow us to have stacked reviews like in gerrit more easily
and send a chain.

We can use front ends for github like https://reviewstack.dev/ (or many
others, but this is a nice one maintained by the folks at facebook) to
see these in a nice way like gerrit provides.


hikalium:
hmm... I think following the normal git workflow here is better to be adopted by more developers. I agree that the Gerrit review interface is nice, but importing it by introducing repo-hooks sounds too much...

I expect your first contribution to be a real code change, not the workflows. You can pick up a thing from the issues. Please check them out and fix one. (I'm happy to help pick one so please let me know offline if you want!)

HidenoriKobayashi:
I agree with @hikalium that this is an overkill for now. I have not seen a case for stacked changes for this project so far. Maybe we could revisit this when such a case arises.

brandonpollack23:
You got it, I just thought we might benefit from this kind of thing since its the workflow we're used to and most contributors will be used to that kind of flow.

For now I'll still use it myself and make small changes that stack but there's no need for everyone to do that.

PR #81, opened 2023-09-29
Update readme to contain stacking information.

brandonpollack23:
Git ps is set up, but we should tell people about it in the readme.


hikalium:
https://github.com/google/lium/pull/80#issuecomment-1740755752

PR #84, opened 2023-10-02
Stacked: Add a rustfmt.toml file with some clean guidlines.

brandonpollack23:
This is based on my own rustfmt.toml I usually use. I tend to set these for the following reasons:

imports_granularity = "Crate" # this combines a bunch of lines at the top, but it doesnt match google style so I removed it
tab_spaces = 2 # param lists and chained calls tend to be longer and more common in rust, I think this increases readability
fn_args_layout = "Tall" # this is way easier to read than non tall
format_strings = true
use_try_shorthand = true # replace any ugly try! macros with the pretty ?
wrap_comments = true
max_width = 100 # If we can do it in Java, we can do it in Rust

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,6 @@
+tab_spaces = 4
+fn_args_layout = "Tall"
+format_strings = true
+use_try_shorthand = true
+wrap_comments = true
+max_width = 100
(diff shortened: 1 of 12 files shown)


brandonpollack23:
ignore, im just figuring out gh command line anf forgot to put draft, sorry for spam

PR #86, opened 2023-10-02
Starter changes

brandonpollack23:
This is a stack of starter changes, each commit works in complete isolation (much like our workflow internally)

I reccomend looking at them one commit at a time (however one does this on github?)

I would love for you to show me how you review this effectively so I can do it too soon :)


src/cmd/setup.rs
```diff
@@ -82,3 +84,18 @@ fn run_bash_completion(_args: &ArgsBashCompletion) -> Result<()> {
     );
     Ok(())
 }
+
+#[derive(FromArgs, PartialEq, Debug)]
+/// Print zsh completion instructions for lium
+#[argh(subcommand, name = "zsh-completion")]
+pub struct ArgsZshCompletion {}
+fn run_zsh_completion(_args: &ArgsZshCompletion) -> Result<()> {
+    run_bash_completion(&ArgsBashCompletion {})?;
```
mhiramat, line 93: This will show the instruction for bash like  "Please run `source  ~/.bash_completion` for the current shell." too.
Can you move common code into a new `install_bash_completion()`  and call it from both `run_bash/zsh_completion()`?

mhiramat (changes requested):

src/cmd.rs
```diff
@@ -33,7 +33,7 @@ pub struct TopLevel {
 
 #[derive(FromArgs, PartialEq, Debug)]
 #[argh(subcommand)]
-/// hikalium's ChromiumOS dev commands
+/// lium's ChromiumOS dev commands
```
mhiramat, line 36: maybe we should keep this :)

mhiramat (approved):
This looks good to me.

src/cmd.rs
brandonpollack23: I understand the sentiment but I think it would cause confusion for external chromiumOS contributors :\

hikalium:
@brandonpollack23 it looks like the check is failing... could you fix that? Thank you!!!

brandonpollack23:
> @brandonpollack23 it looks like the check is failing... could you fix that? Thank you!!!

already about to upload :)

hikalium:
FYI: it looks like the rustfmt is complaining

hikalium:
nice, perfect!!!!!

brandonpollack23:
should be good now.

thanks `git rebase main --exec 'cargo fmt && git commit -a --amend --no-edit'`

PR #91, opened 2023-10-05
Add web page showing DUT console

wtrgjadmw:
The web page which can send commands to DUT via serial port and shows the DUT console is added.

The specification;

1. you can select the serial port which you send commands to from the "select serial" button. After that, the text saying "port is selected" will be shown and the serial port will be opened.
2. you can send commands to your DUT via the port from the form bottom of the web page. Then, the DUT console will be shown on the page. If you didn't select any port, alert window will be shown.

Screenshot 2023-10-05 5 19 59 PM


wtrgjadmw:
Notice;

- DUT default shell is bash, so the console output is decorated. Therefore, the escape sequence is displayed as garbled characters. You should run `/bin/sh` command at first.

docs/servo_power_monitor/prototype/serialport/index.html
hikalium, line 1: The directory structure seems to be too deep...

it may be a good idea to move the file to: `docs/webserial/index.html`

hikalium: Also, when doing so, it's a good idea to add a link in docs/index.md so that users who access to https://google.github.io/lium/ can navigate to your tool!

docs/servo_power_monitor/prototype/serialport/index.html
```diff
@@ -0,0 +1,124 @@
...
+<html>
+
+<head>
+    <meta name="viewport" content="width=device-width,initial-scale=1.0">
+    <title>serial_connection</title>
+    <style>
+        body {
+            margin: 0;
+            padding-bottom: 3rem;
+            font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif;
```
hikalium, line 11: Probably using Mono space font is nice (at least for the console output).

Following pages may help you:
- https://fonts.google.com/?classification=Monospace
- https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#some_common_font_families

docs/servo_power_monitor/prototype/serialport/index.html
```diff
@@ -0,0 +1,124 @@
...
+        #messages>li:nth-child(odd) {
+            background: #efefef;
+        }
+    </style>
+</head>
+
+
+<body>
+    <ul id="messages"></ul>
+    <button id="select">select serial</button>
```
hikalium, line 68: This button should be above the messages / fixed at the bottom with the input box

docs/servo_power_monitor/prototype/serialport/index.html
hikalium: even just using
```
font-family: monospace;
```
will give us a better appearance ;)

docs/servo_power_monitor/prototype/serialport/index.html
```diff
@@ -0,0 +1,124 @@
...
+                return;
+            }
+
+            const chunk = decoder.decode(value, { stream: true });
+            chunk_split_list = chunk.split("\n");
+
+            for (let i = 0; i < chunk_split_list.length - 1; i++) {
+                listItem.textContent += chunk_split_list[i];
+                listItem = document.createElement("li");
+                messages.appendChild(listItem);
```
hikalium, line 103: Can you automatically scroll down to the bottom when new text is arrived...? if you print something long, for example `cat /var/log/messages`, it would be nice

docs/servo_power_monitor/prototype/serialport/index.html
cynthia: font-family: monospace is the standard. Don't pull unnecessary fonts for the time being.

wtrgjadmw:
I fixed reviewed points.

- move directory & add to index.md
- change font
- fix the button position
- scroll automatically when a list item is added

docs/servo_power_monitor/prototype/serialport/index.html
hikalium: it looks like the auto scrolling is not behaving correctly... could you fix that? (maybe that's because the "select serial" button above is taking some space...?)

wtrgjadmw:
I fixed the element sizes and the scrolling bug. Please check when you have time!

hikalium:
LGTM!! Thanks for fixing the issue!

hikalium (approved):

PR #93, opened 2023-10-05
Migrate from printing error logs to using the tokio tracing crate for logging.

brandonpollack23:

### DIFF
--- a/src/cmd/dut.rs
+++ b/src/cmd/dut.rs
@@ -33,6 +33,9 @@ use lium::servo::LocalServo;
 use lium::servo::ServoList;
 use rayon::prelude::*;
 use termion::screen::IntoAlternateScreen;
+use tracing::error;
+use tracing::info;
+use tracing::warn;
 
 #[derive(FromArgs, PartialEq, Debug)]
 /// control DUT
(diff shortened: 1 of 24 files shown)


src/cmd/dut.rs
```diff
@@ -142,16 +146,16 @@ fn run_dut_vnc(args: &ArgsVnc) -> Result<()> {
...
     let mut child_novnc = target.start_port_forwarding(web_port, 6080, "novnc")?;
 
-    eprintln!("To use VNC via web browser, please open:");
-    eprintln!("  http://localhost:{web_port}/vnc.html");
-    eprintln!("To connect via VNC client directly, use:");
-    eprintln!("  xtightvncviewer -encodings raw localhost:{vnc_port}");
+    warn!("To use VNC via web browser, please open:");
+    warn!("  http://localhost:{web_port}/vnc.html");
+    warn!("To connect via VNC client directly, use:");
+    warn!("  xtightvncviewer -encodings raw localhost:{vnc_port}");
```
mhiramat, line 158: This is not a warning message but an information for user to tell how to open VNC.

brandonpollack23:
agreed but I anticipate some users/systems (maybe some time ahead from now)
might filter out "info" and this is important for the user to see so I made
it "warn" level to preserve the message.  As in:

Warning: Just so you know how to do this do that

rather than

Warning: Something is wrong!

brandonpollack23:
lil' ping

src/cmd/dut.rs
hikalium: I think it can be warn to ensure the message is printed out to the users.

hikalium (approved):

PR #96, opened 2023-10-06
Add typescript lint

wtrgjadmw:
[gts](https://github.com/google/gts), google's Typescript style guide and formatter is added to servo_power_monitor directory.


docs/servo_power_monitor/src/index.ts
```diff
@@ -181,7 +187,7 @@ function setupStartUSBButton() {
       });
     } catch (err) {
       console.error(`Disconnected: ${err}`);
-      device = null;
+      // device = null;
       requestUSBButton.disabled = false;
```
wtrgjadmw, line 191: When an error is caught here, nothing is returned in the current implementation because closing usb port without error. But the argument of readLoop is Promise, so TypeScript says the error Argument of type '() => Promise<string | undefined>' is not assignable to parameter of type '() => Promise<string>'. .

cynthia: null and undefined are not the same. You can assign it to undefined and the type check should get through.

wtrgjadmw:
Fixed eslint error.

hikalium (commented):
Looks good to me. Please add proper commit message by editing the first comment in this p-r!

hikalium:
> Looks good to me. Please add proper commit message by editing the first comment in this p-r!

Oh I oversee the pull request's title can be the first line of the commit message... let me merge this with the provided info. sorry @wtrgjadmw for the confision...!

hikalium (approved):

PR #99, opened 2023-10-11
add some instrumentation to capture logs of some child processes

brandonpollack23:


src/util/shell_helpers.rs
```diff
@@ -105,3 +107,43 @@ pub fn run_bash_command_async(cmd: &str, dir: Option<&str>) -> Result<async_proc
...
+    if let Some(ref mut stdout) = child.stdout {
+        for line in std::io::BufReader::new(stdout).lines() {
+            info!("Subprocess {executable} stdout: {}", line.unwrap());
+        }
+    }
+
+    // Read in the stderr of the process, prepend it with an error, and forward it
+    // to our logging.
+    if let Some(ref mut stderr) = child.stderr {
+        for line in std::io::BufReader::new(stderr).lines() {
```
mhiramat, line 141: Can you consolidate the above two loops in one so that `stdout` and `stderr` are read in the same loop? This fixes the order of the output and the error messages and avoid blocking child process on the `stderr` pipe.

mhiramat (changes requested):

src/util/shell_helpers.rs
```diff
@@ -105,3 +107,43 @@ pub fn run_bash_command_async(cmd: &str, dir: Option<&str>) -> Result<async_proc
...
+            info!("Subprocess {executable} stderr: {}", line.unwrap());
+        }
+    }
+
+    child.wait()?;
```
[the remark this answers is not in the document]
brandonpollack23: some callers need access to stdout etc still I believe

src/util/shell_helpers.rs
brandonpollack23: great point, done and added itertools to make it prettier

brandonpollack23:
microping

hikalium (approved):

mhiramat (approved):
This version looks great to me. Thank you for updating!

PR #101, opened 2023-10-11
Add registry to prepare for multiple layers such as indacitif

brandonpollack23:

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -11,7 +11,10 @@ use std::str::FromStr;
 
 use anyhow::Result;
 use tracing::trace;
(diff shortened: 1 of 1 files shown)


hikalium:
Could you explain a bit more about what the pull request does? Thank you!

brandonpollack23:
Yes, so instead of just registering a single subscriber to logs, this creates a registry which can basically manage multiple "layers".  Each layer can have a fileter an do different things depending on the log.  For instance if we want to add progress bars to overall operations that have multiple (parallel) steps etc via indacatif-tracing

brandonpollack23:
itty bitty ping

hikalium (approved):

PR #105, opened 2023-10-17
Draw multiple graphs when dropping JSON files

wtrgjadmw:
This pull request can compare multiple power graphs.

- First, there is no graph (even an empty file)
- When JSON file is dropped, a new graph element is appended and the graph is drawn.
- When Start button is pressed, 
  - if you continue the previous measurement, the graph is drawn on the existing graph.
  - if you start a new measurement, a new graph element is appended and the graph is drawn.


docs/servo_power_monitor/src/index.ts
```diff
@@ -447,7 +447,7 @@ function setupAnalyze() {
   });
 }
 setupAnalyze();
-
+let noGraph = true;
```
hikalium, line 450: this can be moved into setupDataLoad() (to limit the scope of a variable)

PR #108, opened 2023-10-19
Fix for reading subcommand inputs and processing.

brandonpollack23:
With a split in loops, the iterators are always able to continue without
blocking the main thread.


mhiramat (commented):
Just a nitpick comment I added.

mhiramat (approved):
LGTM, Thanks!

PR #111, opened 2023-10-20
110 lium build is stopped

mhiramat:
Revert subprogram logging feature because `lium build` doesn't work correctly with it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -542,15 +542,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "itertools"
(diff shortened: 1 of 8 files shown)


cynthia (approved):

brandonpollack23:
Further context: nothing is technically broken but any process (like portage ones or repo ones) that detect whether they are running in a tty or not will swallow outputs until they complete, making it look like there is no progress and hiding any errors.  This isn't worth the extra logging and any workarounds are complex and fragile

PR #112, opened 2023-10-20
Fix webpack usage to generate a sourcemap correctly

hikalium:
Webpack was not properly used so the generated source maps were for the js files generated by tsc, not the original ts file. Fix it by generating the correct source map with webpack by calling tsc from it, not from the Makefile.

Also add a CI recipe to make sure the typescript sources have no trivial errors by running build and lint.


hikalium:
This change is ready for review. Please check it out!

hikalium:
Note: Generating .js files for GitHub pages via GitHub actions will come in the future. For now, let me allow including the generated code to make it easy...

docs/servo_power_monitor/src/index.ts
```diff
@@ -638,5 +638,6 @@ function setupDataLoad() {
   dropZone.innerText = 'Drop .json here';
   dropZone.addEventListener('dragover', handleDragOver, false);
   dropZone.addEventListener('drop', handleFileSelect, false);
+  console.log("hello!!!!!!!");
```
mhiramat, line 641: Is this a debug message?

hikalium: aaaahhhh, good catch... let me remove that...

docs/servo_power_monitor/src/index.d.ts
```diff
@@ -0,0 +1 @@
+export {};
```
hikalium, line 1: Note: automatically generated by webpack

docs/servo_power_monitor/src/index.ts
hikalium: done! Please review again ;)

mhiramat (approved):
LGTM!

PR #116, opened 2023-10-23
deploy: Switch ~/trunk/ and ~/chromiumos/ in chroot

mhiramat:
Check whether ~/trunk/ exists or not, if not, use ~/chromiumos/ instead for searching update_kernel.sh script.
The `~/trunk` has been removed by https://chromium-review.googlesource.com/c/chromiumos/chromite/+/4835006

### DIFF
--- a/src/cmd/deploy.rs
+++ b/src/cmd/deploy.rs
@@ -72,7 +72,9 @@ pub fn run(args: &Args) -> Result<()> {
             &format!(
                 r###"
 cros-workon-{board} start {packages_str}
-~/trunk/src/scripts/update_kernel.sh {} --remote={} --ssh_port {} --remote_bootargs
+TOPDIR=~/trunk
+[ -d $TOPDIR ] || TOPDIR=~/chromiumos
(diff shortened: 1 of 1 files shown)


src/cmd/deploy.rs
```diff
@@ -72,7 +72,9 @@ pub fn run(args: &Args) -> Result<()> {
             &format!(
                 r###"
 cros-workon-{board} start {packages_str}
-~/trunk/src/scripts/update_kernel.sh {} --remote={} --ssh_port {} --remote_bootargs
+TOPDIR=~/trunk
+[ -d $TOPDIR ] || TOPDIR=~/chromiumos
```
hikalium, line 76: It would be nice if you add a comment that says which CL introduced the dir change (if not and if it seems to take some time to lookup, you can just ignore this comment!)

hikalium (approved):

src/cmd/deploy.rs
mhiramat: Yeah, https://chromium-review.googlesource.com/c/chromiumos/chromite/+/4835006 removes `~/trunk`.

Issue #30, opened 2023-06-26
show repo sync progress while `lium sync` 


hikalium:
Not showing progress on lium sync is not user friendly. display some progress to make it clear that it is not stuck but it is in progress...

brandonpollack23:
https://docs.rs/indicatif/latest/indicatif/ might help with that, along with https://crates.io/crates/tracing-indicatif to integrate it with the logging work in #88

brandonpollack23:
repo sync already outputs a percentage and overwrites it's line using [console control characters](https://man7.org/linux/man-pages/man4/console_codes.4.html), so all we need to do is read the std out of the process, cut out the percentage (using regex or simple string splitting) and turn that into an int as an input for indicatif.  

Then integrate indacatif with the tracing system with the tracing-indicatif crate.

hikalium:
Now it is implemented by @4n12i so closing this as Fixed. Thank you!

PR #117, opened 2023-10-24
Show progress while `lium sync` 

4n12i:
- Get `repo sync` output when running `lium sync`. 
- Print them to stdout.

Issue: #30

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -97,13 +97,12 @@ pub fn repo_sync(repo: &str, force: bool) -> Result<()> {
         let mut buffer = [0; 1];
         let mut stdout = cmd.stdout.take().unwrap();
         loop {
-            let n = stdout.read(&mut buffer)?;
-            if n == 0 {
+            let num_bytes = stdout.read(&mut buffer)?;
+            if num_bytes == 0 {
                 break;
             }
-            let a = format!("ASCII CODE: {:x}", buffer[0]);
-            let cs = std::str::from_utf8(&buffer).unwrap_or(&a);
-            print!("{}", cs);
+            let char = std::str::from_utf8(&buffer)?;
+            print!("{}", char);
         }
 
         let result = cmd
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -81,9 +81,10 @@ pub fn repo_sync(repo: &str, force: bool) -> Result<()> {
         println!("Running repo sync...");
         let repo_sync = format!("repo sync -j{}", &num_cpus::get());
         // `script` is a Unix command that takes a copy of all output to the terminal
-        // and writes it to `typescript` file. Below, explanation of `script`
-        // options. -q Be quiet (do not write start and done messages to
-        // standard output). -e Return the exit status of the child process.
+        // and writes it to `typescript` file.
+        // Below, explanation of `script` options.
+        // -q  Be quiet (do not write start and done messages to standard output).
+        // -e Return the exit status of the child process.
         // -f Flush output after each write.
         // -c Run the command rather than an interactive shell. This makes it easy for a
         // script to capture the output of a program that behaves differently when its
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -96,19 +96,12 @@ pub fn repo_sync(repo: &str, force: bool) -> Result<()> {
             .spawn()
             .context("Failed to execute repo sync")?;
 
-        let mut stdout = cmd
+        let stdout = cmd
             .stdout
             .take()
             .context("Could not get stdout from script output.")?;
-        let mut buffer = [0; 1];
-        loop {
-            let num_bytes = stdout.read(&mut buffer)?;
-            if num_bytes == 0 {
-                break;
-            }
-            let char = std::str::from_utf8(&buffer)?;
-            print!("{}", char);
-        }
+
+        forward_to_std_out(stdout).context("Cannot forward to stdout.")?;
 
         let result = cmd
             .wait_with_output()
@@ -155,6 +148,21 @@ pub fn repo_sync(repo: &str, force: bool) -> Result<()> {
         break;
     }
     println!("repo sync done!");
+    Ok(())
+}
+
+fn forward_to_std_out(mut r: impl Read) -> Result<()> {
+    let mut buffer = [0; 1];
+
+    loop {
+        let num_bytes = r.read(&mut buffer)?;
+        if num_bytes == 0 {
+            break;
+        }
+        let char = std::str::from_utf8(&buffer)?;
+        print!("{}", char);
+    }
+
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


src/repo.rs
```diff
@@ -77,13 +79,33 @@ pub fn repo_sync(repo: &str, force: bool) -> Result<()> {
 
     loop {
         println!("Running repo sync...");
-        let cmd = Command::new("repo")
+        let repo_sync = format!("repo sync -j{}", &num_cpus::get());
+        // `script` is a Unix command that takes a copy of all output to the terminal
```
brandonpollack23, line 83: Interesting.  Are there other places the "repo" command are called that we can do this as well?

If so we should also create a module that handles repo interaction as a whole and does this more generally there and callinto that function from here

src/repo.rs
```diff
@@ -77,13 +79,33 @@ pub fn repo_sync(repo: &str, force: bool) -> Result<()> {
...
+        let mut cmd = Command::new("script")
             .current_dir(repo)
             .stdout(Stdio::piped())
             .stderr(Stdio::piped())
-            .args(["sync", "-j", &num_cpus::get().to_string()])
+            .args(["-qefc", &repo_sync])
             .spawn()
             .context("Failed to execute repo sync")?;
+
+        let stdout = cmd.stdout.take().unwrap();
```
brandonpollack23, line 100: Instead of unwrap I think

```
.context("could not get stdout from script output")?; 
```
would be more appropriate

src/repo.rs
```diff
@@ -77,13 +79,33 @@ pub fn repo_sync(repo: &str, force: bool) -> Result<()> {
...
+        let reader = BufReader::new(stdout);
+        let split_iter = reader
+            .split(b'\r')
```
brandonpollack23, line 103: is it possible to split on either \r or \n?

That way we split on newlines as well and print more readily before the first \r

src/repo.rs
```diff
@@ -77,13 +79,33 @@ pub fn repo_sync(repo: &str, force: bool) -> Result<()> {
...
+            .map(|l| String::from_utf8_lossy(&l.unwrap()).to_string());
+        for a_line in split_iter {
+            print!("{}\r", a_line);
```
brandonpollack23, line 106: I wonder if we should actually replace \r with \n or not?

For user experience I think \r is better but for logs \n is better.

@hikalium what do you think of actually taking a lesson from our sub scirpts and deciding to do one or the other if we are a tty or not?

If we're a tty then \r, if not \n?

src/repo.rs
brandonpollack23: I think repo.rs is a good place for these functions to live.

So you can modify repo.rs to all use the "script" and add this function there as well for syncing.

Wdyt?  (what do you think?)

brandonpollack23 (changes requested):
Great work!  Just some comments to increase clarity/cleanliness and some error handling.

Congrats on your first contribution to lium!

src/repo.rs
```diff
@@ -77,13 +79,36 @@ pub fn repo_sync(repo: &str, force: bool) -> Result<()> {
...
             .spawn()
             .context("Failed to execute repo sync")?;
+
+        let split_iter = BufReader::new(
+            cmd.stdout
+                .take()
+                .context("Could not get stdout from script output.")?,
+        )
+        .split(b'\r')
+        .map(|l| String::from_utf8_lossy(&l.unwrap()).to_string());
```
4n12i, line 106: @brandonpollack23 
Is it better to change to a safer method that does not use `unwrap()` ?

4n12i:
Fixed: 
- Made the process of reading buffer into one sentence for easier code reading.
- Changed error handling to a safer method.

brandonpollack23:
> Fixed:
> 
> * Made the process of reading buffer into one sentence for easier code reading.
> * Changed error handling to a safer method.

Can you reply to the comments where they are stated so we know which ones have been addressed without having to read a general comment?

src/repo.rs
4n12i: I changed it from `unwrap()` to `context()?` .
cec8a0ba935edfb86a9237ad83a16d1d8216c6d8

src/repo.rs
```diff
@@ -77,13 +79,41 @@ pub fn repo_sync(repo: &str, force: bool) -> Result<()> {
 
     loop {
         println!("Running repo sync...");
```
4n12i, line 81: Change the info.

src/repo.rs
4n12i: It isn't possible to separate by one of multiple byte characters using `split()`, so I reverted how to output progress. 
a981ea1080e4110bb3eee1dc55a96768c91840b0

Reference: 
https://doc.rust-lang.org/std/io/trait.BufRead.html#method.split

4n12i: (If we add a progress bar, it may be easier to implement it by splitting the output with `\r`...)

src/repo.rs
```diff
@@ -77,13 +78,38 @@ pub fn repo_sync(repo: &str, force: bool) -> Result<()> {
...
+        let mut buffer = [0; 1];
+        loop {
+            let num_bytes = stdout.read(&mut buffer)?;
+            if num_bytes == 0 {
+                break;
+            }
+            let char = std::str::from_utf8(&buffer)?;
+            print!("{}", char);
+        }
+
```
brandonpollack23, line 112: can you extract (抽出）a function  here called something like:

`fn forward_to_std_out(r: impl Read) -> Result<()>`

brandonpollack23 (changes requested):

src/repo.rs
4n12i: Added a new function which forwards output to stdout!
c8237421edbbda9387d255b1cc584d5d5e03ae9c

4n12i: I have one question: Would it be more readable if I changed these lines to one statement like this?

```rs
forward_to_std_out(
    cmd.stdout
        .take()
        .context("Could not get stdout from script output.")?,
    )
    .context("Could not forward to stdout.")?;
```

brandonpollack23: yeah that seems reasonable to me

brandonpollack23: I think it is fine as is

brandonpollack23 (approved):

src/repo.rs
4n12i: Thank you for your comment!
I updated them.
a403f7956f325cb363e0cc72496515de7e552e2c

src/repo.rs
hikalium: I prefer not changing it implicitly, so how about just going with \n for now...?

hikalium:
LGTM. sorry for the delayed response. Thank you for the contribution!!!

hikalium (approved):

PR #118, opened 2023-10-24
send ctrl+c event to DUT serial port

wtrgjadmw:
In order to cancel any commands in DUT serial port, implemented sending ctrl+c event through input form.


docs/servo_power_monitor/src/index.ts
```diff
@@ -79,14 +79,13 @@ selectDUTSerialButton.addEventListener('click', async () => {
 });
 
 const form = document.getElementById('form') as HTMLFormElement;
+const input = document.getElementById('input') as HTMLInputElement;
```
hikalium, line 82: Please use more descriptive name (but since it affects other lines, you can proceed with this for now)
e.g. DutShellInput

docs/servo_power_monitor/src/index.ts
```diff
@@ -96,6 +95,20 @@ form.addEventListener('submit', async e => {
   }
 });
 
+input.addEventListener('keydown', async e => {
+  if (DUTPort === undefined) {
+    overlay.classList.remove('closed');
+  } else {
```
hikalium, line 101: instead of using else, you can use "early return" to reduce the levels of nesting. so it will looks like:
```
  if (DUTPort === undefined) {
    overlay.classList.remove('closed');
    return;
  }
  const DUTWritable = DUTPort.writable;
```

hikalium (changes requested):

docs/servo_power_monitor/src/index.ts
wtrgjadmw: I fixed it!

wtrgjadmw:
I fixed to "early return"!

docs/servo_power_monitor/src/index.ts
```diff
@@ -79,19 +79,32 @@ selectDUTSerialButton.addEventListener('click', async () => {
...
   } else {
-    const input = document.getElementById('input') as HTMLInputElement | null;
-    if (input === null) return;
     const DUTWritable = DUTPort.writable;
     if (DUTWritable === null) return;
     const DUTWriter = DUTWritable.getWriter();
-    await DUTWriter.write(encoder.encode(input.value + '\n'));
-    input.value = '';
+    if (e.ctrlKey && e.key === 'c') {
+      await DUTWriter.write(encoder.encode('\x03\n'));
```
kalutes, line 106: Can you define the '\x03\n' as a constant value?

docs/servo_power_monitor/src/index.ts
```diff
@@ -79,21 +79,35 @@ selectDUTSerialButton.addEventListener('click', async () => {
...
+input.addEventListener('keydown', async e => {
+  if (DUTPort === undefined) {
+    overlay.classList.remove('closed');
+    return;
+  }
+  const DUTWritable = DUTPort.writable;
+  if (DUTWritable === null) return;
+  const DUTWriter = DUTWritable.getWriter();
+  if (e.ctrlKey && e.key === 'c') {
+    const cancel_cmd = '\x03\n';
```
kalutes, line 107: Not sure about JS style, but in my experience consts are usually defined at the top of the file. Could you move this to the top with the other const (intervalMs)

wtrgjadmw:
defined constant value & move it to the top of `index.ts`

docs/servo_power_monitor/src/index.ts
```diff
@@ -7,6 +7,7 @@ import Dygraph from 'dygraphs';
 import moment from 'moment';
 
 const intervalMs = 100;
+const cancel_cmd = '\x03\n';
```
hikalium, line 10: Please use CONSTANT_CASE instead of snake_case for constants
https://google.github.io/styleguide/tsguide.html#naming-rules-by-identifier-type

hikalium (changes requested):

docs/servo_power_monitor/src/index.ts
wtrgjadmw: should I also fix `intervalMs` to `INTERVAL_MS`?

hikalium: ideally, it should be separated (and the css changes as well), but as far as it is simple enough, okay to change them altogether!

wtrgjadmw: Got it, fixed!

kalutes (approved):

hikalium (approved):
Looks good to me. Thanks!!!
Maybe we can just use the textarea instead of using ul/li to avoid the issue. Please consider that!

PR #127, opened 2023-10-27
delete .eslintrc

### LINKED ISSUE #124: gts doesn't work in /docs/servo_power_monitor
gts (google's typescript lint library) doesn't work in `/docs/servo_power_monitor`.
Also, there are too many `.eslintrc` files and they don't work. 
To solve this issue, it is better to remove `.eslintrc.js` file and change `.eslintrc` file into JSON.

wtrgjadmw:
fix #124


hikalium (approved):
Thanks for catching this!!!

PR #133, opened 2023-10-31
Refactor the process of forwarding output.

4n12i:
Changed from reading output character by character with `Read` to reading all at once with `BufReader` and then creating a byte character iterator.


brandonpollack23 (approved):

mhiramat (approved):
LGTM.

PR #134, opened 2023-10-31
Add a progress bar and verbose option

4n12i:
- Draw a simple progress bar by default while running `lium sync`. 
- With the `--verbose` option, `repo sync` output is printed as it is. 

Screenshot 2023-10-31 4 40 38 PM


mhiramat (approved):
LGTM! Thanks for this feature!

PR #135, opened 2023-10-31
define classes for refactoring

wtrgjadmw:
I fixed the class definition as following;

- `Ui`: it deals with ui components.
- `powerTestController`: it has main functionality in this tool. It saves power data (`powerData`) / connects UI events to the measurement / exports & imports JSON files / analyzes displayed data.
- `ServoController`: it communicates with servo; sends `ina 0` command & reads the output -> parses the power data
- `testRunner`: it runs workload on DUT and reads its serial console.
- `OperatePort`: it is an interface to USB port.
- `graph`, `histogram`: graph drawing


wtrgjadmw:
@hikalium uploaded implementation at the current stage!

wtrgjadmw:
short introduction for classes;

- `powerTestController`: it controls all of this test. 
- `dataRecoder`: I plan it finally corresponds to each iteration (one instance per iteration). but now the user has to stop measurement by hand, so it has almost the same functionality as `powerTestController`.
- `testRunner`: it controls operation on DUT. it runs commands and reads console (finally it also sets up config).
- `operatePort`: it open/close/read/writes from USB/serial port.
- `powerGraph`: it controls graph drawing.
- `histogram`: it controls histogram drawing.

docs/servo_power_monitor/src/dataRecoder.ts
```diff
@@ -0,0 +1,138 @@
...
+    setSerialOutput: (s: string) => void
+  ) {
+    this.servoShell = servoShell;
+    this.enabledRecordingButton = enabledRecordingButton;
+    this.setSerialOutput = setSerialOutput;
+  }
+  changeHaltFlag(flag: boolean) {
+    this.halt = flag;
+  }
+  pushData(s: string) {
```
kawasin73, line 28: pushData() can be refactored to read() method.

```
async read() {
  while (true) {
    const data = await this.servoShell.read();
    this.output += data;

    if (/* is enough data for 1 message */) {
      const message = // parse data
      // clear internal buffer
      this.output = '';
      return message;
    }
  }
}
```

docs/servo_power_monitor/src/dataRecoder.ts
```diff
@@ -0,0 +1,138 @@
+import {powerGraph} from './powerGraph';
+import {operatePort} from './operatePort';
+import {histogram} from './histogram';
+
+export class dataRecoder {
```
kawasin73, line 5: I think we should rename `DataRecorder` to `DataController` or something. This class should focus on talking with servo and message protocol between servo and this monitor.

The important loops (i.e. `kickWriteLoop()`, `readLoop()`) and related database (i.e. `powerData`) and graph should be moved to `PowerTestController` and this class should provide the simple `read()` and `write()` methods.

docs/servo_power_monitor/src/index.ts
```diff
@@ -2,52 +2,152 @@
...
+  dropZone.addEventListener('drop', e => handleFileSelect(e), false);
+  downloadButton.addEventListener('click', () => {
+    const dataStr = controller.exportPowerData();
+    setDownloadAnchor(dataStr);
+  });
+  popupCloseButton.addEventListener('click', () => {
+    overlay.classList.add('closed');
+  });
+
+  function enabledRecordingButton(halt: boolean) {
```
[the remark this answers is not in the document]
wtrgjadmw: does it mean all functions related to UI components (L85~L154) should be moved to a new class? If so, should `.getElementbyId` and `.addEventLister` also be moved from index.ts to the class?

docs/servo_power_monitor/src/dataRecoder.ts
wtrgjadmw: I think simple `read()` and `write()` methods are in `OperatePort` class, so I renamed `DataRecoder` to `DataParser` and made the class focus on storing servo's output + extract the power data.

wtrgjadmw:
I fixed the class definition as following;

- `powerTestController`: it has main functionality in this tool. It saves power data (`powerData`) / connects UI events to the measurement / exports & imports JSON files / analyzes displayed data.
- `ServoController`: it communicates with servo; sends `ina 0` command & reads the output -> parses the power data
- `testRunner`: it runs workload on DUT and reads its serial console.
- `OperatePort`: it is an interface to serial / USB port.
- `graph`, `histogram`: graph drawing

wtrgjadmw:
I reflected the removal of webUSB!

docs/servo_power_monitor/src/index.ts
```diff
@@ -1,46 +1,96 @@
...
-  executeScriptAddClickEvent,
-  formAddSubmitEvent,
-  haltAddClickEvent,
-  inputAddKeydownEvent,
-  requestSerialAddClickEvent,
-  selectDutSerialAddClickEvent,
-  setPopupCloseButton,
-} from './ui';
+import moment from 'moment';
+import {OperatePort} from './operatePort';
```
hikalium, line 5: https://google.github.io/styleguide/tsguide.html#identifiers-imports

File names should be snake_case

docs/servo_power_monitor/src/ui.ts
```diff
@@ -1,103 +1,61 @@
...
+  public dutCommandForm = document.getElementById(
+    'dutCommandForm'
+  ) as HTMLFormElement;
+  public dutCommandInput = document.getElementById(
+    'dutCommandInput'
+  ) as HTMLInputElement;
+  public popupCloseButton = document.getElementById(
+    'popup-close'
+  ) as HTMLButtonElement;
+  public overlay = document.querySelector('#popup-overlay') as HTMLDivElement;
```
hikalium, line 26: Maybe using getElementById instead is a good idea for the consistency

hikalium (changes requested):
Except for two comments looks good!!!

wtrgjadmw:
Fixed!

hikalium (approved):

PR #137, opened 2023-11-08
remove webUSB

wtrgjadmw:
This PR removes "Start (webUSB)" button to make the codes simpler.
It was implemented to support MacOS, which doesn't support communication via WebSerial. But for now, we are focusing on ChromeOS as a workstation and DUT, so it can be removed.
If MacOS is needed for the future, you can reproduce the codes from git log.


hikalium (approved):
LGTM!

mblsha:
Seems that the message needs to be updated, since webUSB is removed.
Screenshot 2023-11-15 at 12 02 24 pm

mblsha:
Turns out after this change it won't work on macOS anymore since it doesn't have a driver that would expose Servo's serial ports in a way that would be visible by Chrome.

wtrgjadmw:
The tool aims to support the power consumption measurement of a Chromebook for now, so we deleted this to avoid the code structure becoming too complicated. But it is still future work to support webUSB connection again.

cynthia:
I personally think this should be reverted and reworked so that it does WebSerial in ChromeOS/Linux and WebUSB elsewhere. I suspect the Windows path also needs WebUSB, which is likely what our partners would be using.

hikalium:
Thanks for the feedback everyone! Created #159 to revive the WebUSB support. Please give us some time as our workload is limited. (if you want to work on it, please let me know so that I can re-assign the issue to you!!)

PR #138, opened 2023-11-09
Add annotation in power graph

wtrgjadmw:
- add annotation in power graph when the "start" or "end" signal comes in  DUT console like the below image
- x value is expressed as a relative time from the start of measurement
- to avoid adding annotation when copying shell script to DUT, encode script by base64
- include annotation information to exported JSON file. The data format is defined in [design doc](https://docs.google.com/document/d/1kmsSz7AaATIMxwoVpHFu-ZsWusrfRd7UARaf2bJmjMo/edit?resourcekey=0-7Y93jZmJWD-EItWLC03VzQ#heading=h.p958ooeyscy7)

image


wtrgjadmw:
changed annotation style!

PR #140, opened 2023-11-09
Add annotation

wtrgjadmw:
- add annotation in power graph when the "start" or "end" signal comes in  DUT console like the below image
- x value is expressed as a relative time from the start of measurement
- to avoid adding annotation when copying shell script to DUT, encode script by base64
- include annotation information to exported JSON file. The data format is defined in [design doc](https://docs.google.com/document/d/1kmsSz7AaATIMxwoVpHFu-ZsWusrfRd7UARaf2bJmjMo/edit?resourcekey=0-7Y93jZmJWD-EItWLC03VzQ#heading=h.p958ooeyscy7)

image


docs/servo_power_monitor/src/graph.ts
```diff
@@ -1,26 +1,39 @@
 import Dygraph from 'dygraphs';
 import {Ui} from './ui';
+import {AnnotationText, PowerData} from './power_test_controller';
 
 export class Graph {
   private ui: Ui;
-  private g = new Dygraph('graph', [], {});
+  public g = new Dygraph('graph', [], {});
+  public annotations;
```
hikalium, line 8: please add initial value ` = [];`

docs/servo_power_monitor/src/graph.ts
```diff
@@ -1,26 +1,39 @@
...
+  public annotationFlag = false;
+  private annotationText: AnnotationText = 'start';
   constructor(ui: Ui) {
     this.ui = ui;
+    this.annotations = this.g.annotations();
```
hikalium, line 13: Do we have to keep the previous annotations exists in the graph `this.g`? As the graph is created at the constructor, it should be empty, so this.annotations can just be an empty array (so we don't need this line.)

docs/servo_power_monitor/src/graph.ts
```diff
@@ -1,26 +1,39 @@
...
+        xlabel: 'Relative Time (s)',
         ylabel: 'Power (mW)',
         legend: 'always',
         showRangeSelector: true,
         connectSeparatedPoints: true,
+        axes: {
+          x: {
+            axisLabelFormatter: function (d) {
+              const relativeTime = (d as number) - powerData[0][0];
+              return (relativeTime / 1000).toLocaleString();
```
hikalium, line 33: It would be nice if there is a comment that describes why the time is divided by 1000 (I guess since the time recorded is in ms but the unit of axis is s)

docs/servo_power_monitor/src/graph.ts
```diff
@@ -35,6 +48,45 @@ export class Graph {
...
     );
+    if (this.annotationFlag) {
+      this.addAnnotation(
+        powerData[powerData.length - 1][0],
+        this.annotationText
+      );
+      this.annotationFlag = false;
+    }
+  }
+  public setAnnotationFlag(text: AnnotationText) {
```
hikalium, line 59: I think this logic (which delays adding an annotation on the next update) should be outside of graph.ts, and the graph only provides a way to set annotation on the last data point. Let's discuss later.

hikalium (changes requested):

docs/servo_power_monitor/src/graph.ts
```diff
@@ -36,6 +47,34 @@ export class Graph {
...
+    };
+    this.annotations.push(newAnnotation);
+    this.g.setAnnotations(this.annotations);
+  }
+  public findAnnotationPoint(
+    powerData: Array<PowerData>,
+    time: number,
+    text: string
+  ) {
+    for (const powerDataElement of powerData) {
```
hikalium, line 71: Maybe walking over the elements in a reversed order and finding the first element that is earlier than the annotation point would be faster than this logic

docs/servo_power_monitor/src/graph.ts
```diff
@@ -1,26 +1,37 @@
...
+        xlabel: 'Relative Time (s)',
         ylabel: 'Power (mW)',
         legend: 'always',
         showRangeSelector: true,
         connectSeparatedPoints: true,
+        axes: {
+          x: {
+            axisLabelFormatter: function (d) {
+              const relativeTime = (d as number) - powerData[0][0];
+              // relativeTime is divided by 1000 because the time data is recorded in miliseconds but x-axis is in seconds.
```
hikalium, line 30: nit: milliseconds (double L)

hikalium (commented):
LGTM except for a few points!

hikalium (approved):

PR #142, opened 2023-11-13
Make shell scripts customizable

wtrgjadmw:
add textarea and set its content as workload when "start" button is pressed

(have not checked on local servo yet)


hikalium (approved):
LGTM

PR #143, opened 2023-11-14
run workload & stop the measurement automatically

wtrgjadmw:
- after "Start" button is pressed, the tool starts measurement and runs workload on DUT
- when workload finishes, the measurement stops automatically.
- delete "select DUT Port", "Execute" button & change "Stop" to "Cancel"

(have not checked on local servo yet)


docs/servo_power_monitor/src/test_runner.ts
```diff
@@ -2,26 +2,39 @@ import {OperatePort} from './operate_port';
...
   private CANCEL_CMD = '\x03\n';
+  private isOpened = false;
   // shell script
   private scripts = `#!/bin/bash -e
 function workload () {
-  stress-ng -c 1 -t $1
+  stress-ng -c 1 -t 10
 }
-ectool chargecontrol idle
+ectool chargecontrol idle 1> ./test_out.log 2> ./test_err.log
```
hikalium, line 12: How about wrapping those top-level commands that their outputs are redirected into the same file as a function? In that way, we can avoid repeating redirects. e.g.
```
function run_iteration () {
  ectool chargecontrol idle
  workload
  ...
}
run_iteration
```
FYI: there is a Shell style guide: https://google.github.io/styleguide/shellguide.html#s7-naming-conventions

Also, it may be a good idea to refactor the ectool chargecontrol calls and/or put comments on them since it's not clear what they are doing for the people who don't know about ectool well...

wtrgjadmw: I agree that repeating redirects is redundant, but I think ectool chargecontrol should be called only once during one measurement. So how about executing the command outside of the script, such as before the cat command?

docs/servo_power_monitor/src/power_test_controller.ts
```diff
@@ -59,9 +59,8 @@ export class PowerTestController {
     }
   }
   private async readDutLoop() {
-    this.runner.executeCommand('\n');
-    this.ui.addMessageToConsole('DutPort is selected');
-    for (;;) {
+    // await this.runner.executeCommand('\n');
```
kalutes, line 62: uncomment or remove

docs/servo_power_monitor/src/test_runner.ts
hikalium: Yeah that seems to be a better solution! Could you update the code to do so?

wtrgjadmw: already updated! please check L33, L37 in `test_runner.ts`

docs/servo_power_monitor/src/power_test_controller.ts
```diff
@@ -75,26 +73,29 @@ export class PowerTestController {
...
+    await this.changeHaltFlag(false);
+    this.readDutLoop();
     this.kickWriteLoop();
     this.readLoop();
+    await this.runner.copyScriptToDut();
+    await this.runner.executeScript();
   }
-  public async stopMeasurement() {
+  public async stopMeasurement(abortFlag = false) {
+    if (abortFlag) await this.runner.sendCancel();
```
hikalium, line 94: Does the function need an abort flag? How about just calling sendCancel() when this.inProgress is true at the beginning of this function?

wtrgjadmw: I added this `sendCancel()` to finish the running command on DUT console, but `this.inProgress` and `this.halt` flag control the read/write loop on servo. So I think whether `this.inProgress` is true or false doesn't matter. However, I agree that it doesn't `abortFlag` and we can just call `sendCancel()` regardless of whether some command is running or not on DUT.

hikalium: I'm not sure whether `await this.runner.sendCancel()` finishes even when the case which was abortFlag == false. Did you check that case...? (I'm just curious and it may be a rare error case so it may not be a huge matter but I'm curious if you checked that. thanks!!)

hikalium (approved):
LGTM. nice job!

docs/servo_power_monitor/src/power_test_controller.ts
hikalium: Discussed with @wtrgjadmw offline, and we decided to go with the current implementation (call sendCancel anyway even if it is already closed) since it seems to be okay for now and simplify the implementation.

docs/servo_power_monitor/src/test_runner.ts
hikalium: thanks!! checked!

PR #144, opened 2023-11-14
Add the number of config

wtrgjadmw:
- add "Add config" button and "configList" to control the number & the content of config.
- run workload for each config alternately.
- load & export JSON with the multiple config (ref: [design doc](https://docs.google.com/document/d/1kmsSz7AaATIMxwoVpHFu-ZsWusrfRd7UARaf2bJmjMo/edit?resourcekey=0-7Y93jZmJWD-EItWLC03VzQ#heading=h.p958ooeyscy7))


docs/servo_power_monitor/index.html
```diff
@@ -14,14 +14,20 @@
   </div>
   <div id="controlDiv">
     <a id="downloadAnchorElem" style="display:none"></a>
-    <button id='requestSerialButton'>Start</button>
+    <button id='requestSerialButton'>Start(WebSerial)</button>
+    <button id='request-device'>Start(WebUSB)</button>
```
kalutes, line 18: I think webusb support was removed, right?

docs/servo_power_monitor/src/index.ts
```diff
@@ -51,17 +44,9 @@ window.addEventListener('DOMContentLoaded', () => {
...
-  });
-  ui.executeScriptButton.addEventListener('click', async () => {
-    if (!runner.isOpened) {
-      ui.overlay.classList.remove('closed');
-      return;
-    }
-    await runner.copyScriptToDut();
-    await runner.executeScript();
-  });
+  // ui.analyzeButton.addEventListener('click', () => {
```
kalutes, line 47: Uncomment or remove.

docs/servo_power_monitor/index.html
```diff
@@ -14,14 +14,20 @@
...
     <button id='selectDutSerialButton'>Select DUT Port</button>
     <button id='executeScriptButton'>Execute scripts</button>
     <button id='haltButton'>Stop</button>
     <button id='downloadButton'>Download</button>
     <button id='downloadButton'>Mark</button>
     <button id='analyzeButton'>Analyze displayed range</button>
   </div>
+  <div id="config">
+    <button id="addConfigButton">Add config</button>
```
kalutes, line 27: Instead of having "Add Config" and "Delete Config" buttons, it's probably more user friendly to have a single "Add Config" button, and separate "Delete" buttons for each individual config.

kalutes: This way the user can, for example, delete a config in the middle of the list.

docs/servo_power_test/index.css
```diff
@@ -127,6 +131,10 @@ body {
   position: relative;
 }
 
+#shellScriptList {
```
hikalium, line 134: I think we should use kebab-case here for the ids.
https://google.github.io/styleguide/htmlcssguide.html#id_Attributes

docs/servo_power_test/index.css
```diff
@@ -163,32 +171,27 @@ body {
...
-  margin-bottom: 2px;
-}
-
 #d3area {
   width: 50%;
   overflow-y: scroll;
 }
 
-#console {
+#dut_console {
```
hikalium, line 179: (ditto) I think we should use kebab-case here for the ids.
https://google.github.io/styleguide/htmlcssguide.html#id_Attributes

docs/servo_power_test/index.html
```diff
@@ -17,29 +17,24 @@
     <button id='requestSerialButton'>Start</button>
     <button id='haltButton'>Cancel</button>
     <button id='downloadButton'>Download</button>
-    <button id='analyzeButton'>Analyze displayed range</button>
```
hikalium, line 20: Maybe we should do that in the later commit, but the id should be kebab-case (sorry I should have notice that earlier...) https://google.github.io/styleguide/htmlcssguide.html#id_Attributes

docs/servo_power_test/src/config.ts
```diff
@@ -0,0 +1,107 @@
...
+  private INTERVAL_MS = 100;
+  private ui: Ui;
+  private servoController: ServoController;
+  private runner: TestRunner;
+  public powerDataList: Array<PowerData> = [];
+  public annotationList: Array<AnnotationData> = [];
+  public graph: Graph;
+  public customScript: string;
+  public halt = true;
+  private inProgress = false;
```
hikalium, line 17: sort them with visibility (private / public)

docs/servo_power_test/src/config.ts
```diff
@@ -0,0 +1,107 @@
...
+export class Config {
+  private INTERVAL_MS = 100;
+  private ui: Ui;
+  private servoController: ServoController;
+  private runner: TestRunner;
+  public powerDataList: Array<PowerData> = [];
+  public annotationList: Array<AnnotationData> = [];
+  public graph: Graph;
+  public customScript: string;
+  public halt = true;
```
hikalium, line 16: do we have to make the halt flag public...? less public fields are generally better.
https://google.github.io/styleguide/tsguide.html#visibility

wtrgjadmw:
@hikalium As we discussed at 1:1, I'm planning to modify the annotationList structure to dict in Config class to dict, but should I do it in this PR? Or separate?

docs/servo_power_test/src/config.ts
```diff
@@ -0,0 +1,107 @@
...
+      }
+    }
+  }
+  public async start() {
+    await this.runner.openDutPort();
+    await this.servoController.openServoPort();
+    await this.changeHaltFlag(false);
+    this.kickWriteLoop();
+    this.readLoop();
+    const readDutLoopInst = this.readDutLoop();
```
hikalium, line 94: maybe `readDutLoopPromise` is better (`Inst` is I guess shortened version of `Instance`, which should be avoided as the following docs say:)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function
https://google.github.io/styleguide/tsguide.html#descriptive-names

docs/servo_power_test/src/graph.ts
```diff
@@ -4,10 +4,14 @@ import {AnnotationData, PowerData} from './power_test_controller';
 
 export class Graph {
   private ui: Ui;
-  public g = new Dygraph('graph', [], {});
+  public g: Dygraph;
```
hikalium, line 7: consider if we can make it private (as mentioned in the above)

docs/servo_power_test/src/graph.ts
```diff
@@ -4,10 +4,14 @@ import {AnnotationData, PowerData} from './power_test_controller';
...
   public annotations: dygraphs.Annotation[] = [];
-  constructor(ui: Ui) {
+  constructor(ui: Ui, divName: string) {
```
hikalium, line 9: I think passing HtmlDivElement instead of string of the div's id is better (as Dygraph takes both of them):
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/dygraphs/index.d.ts#L1160

hikalium (changes requested):

hikalium:
> @hikalium As we discussed at 1:1, I'm planning to modify the annotationList structure to dict in Config class to dict, but should I do it in this PR? Or separate?

Whichever is fine :) but I think this pull request has a lot of things to do so maybe being separated is better...

docs/servo_power_test/index.css
wtrgjadmw: there are too many lowerCamelCase id, so I will fix only elements which are added in this PR & fix other elements in another PR.

wtrgjadmw: (we also have to fix the ids in servo_power_monitor directory...)

hikalium: sounds good, yeah that would be a good choice. thank you!

hikalium (approved):
LGTM!

PR #146, opened 2023-11-15
Create .gitattributes to hide generated files from the review by default

hikalium:
Mark docs/servo_power_monitor/dist/* as generated files to make the review easier. The js files under the dir should be kept checked-in for a while since it is used for github pages (will be replaced by the actions later)


HidenoriKobayashi (approved):
I just checked that your change matches the description of this PR.

PR #150, opened 2023-11-16
add servo_power_test page

wtrgjadmw:
duplicate `servo_power_monitor` directory and rename it as `servo_power_test`.


hikalium:
Could you modify `https://github.com/google/lium/blob/main/.gitattributes` as well? This file prevents generated files under dist/ dir not to show up on the PR review diffs.

.github/workflows/typescript.yml
```diff
@@ -16,8 +16,8 @@ jobs:
         node-version: '18.x'
     - uses: actions/checkout@v4
     - name: Build
-      working-directory: ./docs/servo_power_monitor/
+      working-directory: ./docs/servo_power_test/
       run: npm install
     - name: Run tests
-      working-directory: ./docs/servo_power_monitor/
+      working-directory: ./docs/servo_power_test/
```
hikalium, line 22: I think we have to duplicate the github actions workflow for both of the tools (servo_power_monitor, servo_power_test). Please update the YAML file to do tests on both of them 🙏

.gitattributes
```diff
@@ -1 +1 @@
-docs/servo_power_monitor/dist/* linguist-generated=true
+docs/servo_power_test/dist/* linguist-generated=true
```
hikalium, line 1: How about keep both of them, since the old one (monitor) will be there still...

docs/servo_power_test/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "cr50",
```
hikalium, line 2: could you update the name to match with each project? (Also you can update "author" attribute as well if you want ;) )

hikalium (changes requested):

hikalium (approved):

PR #153, opened 2023-11-17
Force deploy to avoid failure caused by USE mismatch warnings

hikalium:
Force deploy when running `cros deploy` via `lium deploy`. `cros deploy` started checking if USE flags are matched when deploying and building, and when it is not matched, it asks users to proceed or not (and... the command itself will succeed even if the stdin sends just an EOF, which is not good...), which causes a false success on the lium side. To avoid this, for now, calling `cros deploy` with `--force` flag to continue without that prompt.

### DIFF
--- a/src/cmd/deploy.rs
+++ b/src/cmd/deploy.rs
@@ -128,7 +128,7 @@ fn cros_workon_user_packages(
         chroot.run_bash_script_in_chroot(
             "deploy",
             &format!(
-                r"cros-workon-{board} start {packages_str} && cros deploy {} {user_pkgs}",
+                r"cros-workon-{board} start {packages_str} && cros deploy --force {} {user_pkgs}",
                 target.host_and_port()
             ),
             None,


hikalium:
Output before the change:
```
(workstation) $ lium deploy --repo . --dut ${DUT} crosvm arcvm-kernel-ack-5_10                                                    
...                                                                                                 
12:33:08: WARNING: USE flags for package sys-kernel/arcvm-kernel-ack-5_10 do not match (Existing='abi_x86_64 amd64 android-vm-rvc arcvm board_use_r
ammus builtin_fw_x86_aml_ucode cheets chromeos_kernel_family_chromeos clang dt_compression elibc_glibc fit_compression_kernel_lz4 iioservice kernel
_linux lld llvm_ias tpm2 unibuild userland_GNU vdso32', New='abi_x86_64 amd64 android-vm-rvc arcvm board_use_rammus builtin_fw_x86_aml_ucode cheets
 chromeos_kernel_family_chromeos clang dt_compression elibc_glibc fit_compression_kernel_lz4 iioservice kernel_linux lld llvm_ias pcserial tpm2 uni
build userland_GNU vdso32').                                                                                                                       
12:33:08: NOTICE: These are the packages to emerge:                                                                                                
12:33:08: NOTICE: * 1) chromeos-base/crosvm-9999 (root=/)                                                                                          
12:33:08: NOTICE: * 2) sys-kernel/arcvm-kernel-ack-5_10-9999 (root=/)                                                                              
12:33:08: WARNING: Continue despite prior warnings?                                                                                                
                                                                                                                                                   
Do you want to continue? (yes/No)?                                                                                                                 
12:33:08: NOTICE: cros deploy completed successfully in 14.711s                                                                                    
Rebooting DUT...
```

HidenoriKobayashi (approved):
I took this as a temporary workaround. Since there may be a case where a USE flag mismatch would cause a surprising result, I'd like at least some sort of a warning message to be emitted. Is there a new issue to properly handle this?

hikalium:
Yeah, it should be a temporarily thing, so I created #158 as an action item.
It is also possible that if the build system becomes non-sysroot based (stateless), the issue may goes away since the only part the USE flag has some effect will be the build phase. Anyway, thanks for the review!

PR #157, opened 2023-11-21
Modify annotationList structure to dict

wtrgjadmw:
Modified the structure of annotationList in Config class to Map from array in order to make it easier to get the time when workload start/ends running.
Object.fromEntries() is supported from node.js 12.0.0, so this PR also updated node.js in package.json.


docs/servo_power_test/src/power_test_controller.ts
```diff
@@ -4,7 +4,7 @@ import {TestRunner} from './test_runner';
 import {Config} from './config';
 
 export type PowerData = [number, number];
-export type AnnotationData = [number, string];
+export type AnnotationDataList = {[labelName: string]: number};
```
hikalium, line 7: Maybe using ES6's Map instead of using Object is better.
c.f. https://google.github.io/styleguide/tsguide.html#indexable-key-string-type

hikalium (changes requested):

docs/servo_power_test/src/power_test_controller.ts
wtrgjadmw: it seems that Map cannot be converted to JSON directly. https://typescriptbook.jp/reference/builtin-api/map#map%E3%81%AF%E7%9B%B4%E6%8E%A5json%E3%81%AB%E3%81%A7%E3%81%8D%E3%81%AA%E3%81%84
we have to create another variable, so it looks using extra memory. but is it okay...?

hikalium: It's not a memory-critical part and I think it would be better to take the benefits of type-safe (and less-magic-of-javascript) structure (Map).

wtrgjadmw: I got it!

docs/servo_power_test/package.json
```diff
@@ -33,6 +33,6 @@
     "@types/node": "20.8.2"
   },
   "engines": {
-    "node": ">= 11.0.0"
+    "node": ">= 12.0.0"
```
hikalium, line 36: nit: is it intended? (just making sure...)

hikalium (approved):

docs/servo_power_test/package.json
wtrgjadmw: Object.fromEntries() is supported from node.js 12.0.0, so updated node.js in package.json.

hikalium: Got it. Thanks for the explanation!!

PR #163, opened 2023-11-22
Run workload for multiple iterations

wtrgjadmw:
- after the start button is pressed, it runs workload for each config alternately for multiple times
- add input area to define the number of iterations
- add selector to show the selected iteration's graph
- define IterationData to save powerData and annotationData for each iteration
- with the change of saved-data structure, the JSON file design is also changed ([design doc](https://docs.google.com/document/d/1kmsSz7AaATIMxwoVpHFu-ZsWusrfRd7UARaf2bJmjMo/edit?resourcekey=0-7Y93jZmJWD-EItWLC03VzQ#heading=h.p958ooeyscy7))


docs/servo_power_test/index.html
```diff
@@ -17,6 +17,7 @@
     <button id='request-serial-button'>Start</button>
     <button id='halt-button'>Cancel</button>
     <button id='download-button'>Download</button>
+    <span>number of iterations: <input id="iteration-input" type="number" placeholder="1" style="width: 32px;"></span>
```
hikalium, line 20: I think all usage of `style` attribute should be in .css files to separate the concerns, as said in the style guide:
https://google.github.io/styleguide/htmlcssguide.html#Separation_of_Concerns

docs/servo_power_test/index.html
```diff
@@ -27,12 +28,17 @@
...
     <span id="dut-console"></span>
   </div>
-  <div id="popup-overlay" class="closed">
+  <div id="popup-overlay" class="hidden">
     <div id="popup-window">
       <p id="popup-text">DUT serial port is not selected</p>
       <button id="popup-close">close</button>
     </div>
   </div>
+  <div style="display: flex;">
```
hikalium, line 37: ditto

docs/servo_power_test/src/config.ts
```diff
@@ -4,17 +4,29 @@ import {ServoController} from './servo_controller';
...
 export class Config {
   private INTERVAL_MS = 100;
   private ui: Ui;
   private servoController: ServoController;
   private runner: TestRunner;
   private halt = true;
   private inProgress = false;
-  public powerDataList: Array<PowerData> = [];
-  public annotationList: AnnotationDataList = new Map<string, number>();
+  public iterationDataList: Array<IterationData> = [];
```
hikalium, line 26: Can you add methods to access iterationDataList to make it private? Less public fields is a good practice, if there are no strong reasons.
c.f. https://google.github.io/styleguide/tsguide.html#visibility

hikalium (changes requested):

hikalium (approved):
LGTM. Thanks!

wtrgjadmw:
rebased & resolved merge conflict

docs/servo_power_test/src/config.ts
```diff
@@ -4,17 +4,41 @@ import {ServoController} from './servo_controller';
 import {TestRunner} from './test_runner';
 import {Ui} from './ui';
 
+export class IterationData {
+  public powerDataList: Array<PowerData>;
+  public annotationList: AnnotationDataList;
```
hikalium, line 9: Can you make those two properties private? doing a similar thing (implementing getter / setter) as extractData does will make the encapsulation better ;)

docs/servo_power_test/src/config.ts
```diff
@@ -4,17 +4,41 @@ import {ServoController} from './servo_controller';
...
   private servoController: ServoController;
   private runner: TestRunner;
   private halt = true;
   private inProgress = false;
-  public powerDataList: Array<PowerData> = [];
-  public annotationList: AnnotationDataList = new Map<string, number>();
+  private iterationDataList: Array<IterationData> = [];
   public graph: Graph;
   public customScript: string;
+  public currentItrNum = 0;
```
hikalium, line 41: Could you make this private...? (without an explicit reason, all the class properties should be private.)

docs/servo_power_test/src/power_test_controller.ts
```diff
@@ -15,6 +15,7 @@ export class PowerTestController {
   private totalHistogram = new TotalHistogram();
   private configList: Array<Config> = [];
   private currentConfigNum = 0;
+  private itrNum = 2;
```
hikalium, line 18: Also, maybe "itr" should be "iteration". A descriptive name is preferred!
https://google.github.io/styleguide/tsguide.html#descriptive-names

docs/servo_power_test/src/power_test_controller.ts
```diff
@@ -48,43 +49,48 @@ export class PowerTestController {
...
+    if (this.itrNum <= 0) return;
     await this.servoController.servoShell.select();
     await this.runner.dut.select();
     await this.initializePort();
     await this.setConfig();
-    for (let i = 0; i < this.ui.configNum; i++) {
-      this.currentConfigNum = i;
-      console.log(`start running config${i}`);
-      await this.configList[i].start();
+    for (let currentItrNum = 0; currentItrNum < this.itrNum; currentItrNum++) {
```
hikalium, line 58: while the local variables should have a descriptive name, maybe single or two character name here is good for readability.
> Exception: Variables that are in scope for 10 lines or fewer, including arguments that are not part of an exported API, may use short (e.g. single letter) variable names.
https://google.github.io/styleguide/tsguide.html#descriptive-names

docs/servo_power_test/src/power_test_controller.ts
```diff
@@ -48,43 +49,48 @@ export class PowerTestController {
...
-        ) {
-          extractedData.push(powerData[1]);
-        }
-      }
+      const extractedData = config.extractTotalHistogramData(this.marginTime);
       histogramData.push(extractedData);
     }
     this.totalHistogram.paintHistogram(histogramData);
   }
+  public showSelectedItrGraph(selectedItr: number) {
```
hikalium, line 82: Itr -> Iteration

docs/servo_power_test/src/ui.ts
```diff
@@ -78,7 +87,19 @@ export class Ui {
...
   }
-  public hideToolTip() {
-    this.toolTip.classList.add('hidden');
+  public hideElement(element: HTMLElement) {
+    element.classList.add('hidden');
+  }
+  public showElement(element: HTMLElement) {
+    element.classList.remove('hidden');
+  }
+  public appendItrSelectors(itrNum: number, selectedIndex: number) {
```
hikalium, line 96: Itr -> Iteration

hikalium (changes requested):

docs/servo_power_test/src/ui.ts
```diff
@@ -78,7 +87,19 @@ export class Ui {
...
+    element.classList.remove('hidden');
+  }
+  public appendItrSelectors(iterationNumber: number, selectedIndex: number) {
+    for (let i = 0; i < iterationNumber; i++) {
+      const newOption = document.createElement('option');
+      newOption.innerText = `${i + 1}`;
+      this.iterationSelector.add(newOption);
+    }
+    this.iterationSelector.selectedIndex = selectedIndex;
+    this.showElement(this.iterationSelector);
```
hikalium, line 103: Is the iteration selector hidden by default so the showElement call is needed here? If so, adding a comment that explains so would help the readers to understand the code ;) (since show and hide is mostly paired usually, I was curious why it only calls show here...)

hikalium:
It's almost there (good work!), but I left some comments, so please check that! :)

docs/servo_power_test/src/config.ts
```diff
@@ -4,32 +4,112 @@ import {ServoController} from './servo_controller';
 import {TestRunner} from './test_runner';
 import {Ui} from './ui';
 
+export class IterationData {
+  private powerDataList: Array<PowerData>;
+  private annotationList: AnnotationDataList;
+  private graph: Graph;
```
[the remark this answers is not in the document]
wtrgjadmw: I agree that the current implementation is redundant (especially in config.ts L23~L25, L33~35), but one graph is corresponding to one iteration, so I think it's not bad.

docs/servo_power_test/src/ui.ts
wtrgjadmw: you're right. Added a comment.

docs/servo_power_test/src/config.ts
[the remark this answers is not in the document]
hikalium: Ah, that's right... then I think it's okay. Thank for the explanation!

hikalium (approved):

PR #168, opened 2023-11-22
92 list known board and complete it

mhiramat:
Add `lium board list` and use it for completion.

### DIFF
--- a/src/cmd/board.rs
+++ b/src/cmd/board.rs
@@ -45,7 +45,7 @@ pub struct ArgsList {
     #[argh(positional)]
     filter: Option<String>,
 
-    /// only show cached list
+    /// only show cached data without updating it (fast)
     #[argh(switch)]
     cached: bool,
 }
--- a/src/cmd/board.rs
+++ b/src/cmd/board.rs
@@ -1,0 +1,92 @@
+// Copyright 2023 The ChromiumOS Authors
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+use anyhow::bail;
+use anyhow::Result;
+use argh::FromArgs;
+use glob::Pattern;
+use lium::cache::KvCache;
+use lium::repo::get_repo_dir;
+use lium::util::shell_helpers::run_bash_command;
+
+#[derive(FromArgs, PartialEq, Debug)]
+/// list supported boards
+#[argh(subcommand, name = "board")]
+pub struct Args {
+    #[argh(subcommand)]
+    nested: SubCommand,
+}
+
+pub static BOARD_CACHE: KvCache<Vec<String>> = KvCache::new("board_cache");
+
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand)]
+enum SubCommand {
+    List(ArgsList),
+}
+pub fn run(args: &Args) -> Result<()> {
+    match &args.nested {
+        SubCommand::List(args) => run_board_list(args),
+    }
+}
+
+#[derive(FromArgs, PartialEq, Debug)]
+/// List up the supported boards
+#[argh(subcommand, name = "list")]
+pub struct ArgsList {
+    /// target cros repo directory
+    #[argh(option)]
+    repo: Option<String>,
+
+    /// glob pattern of the boards
+    #[argh(positional)]
+    filter: Option<String>,
+
+    /// only show cached list
+    #[argh(switch)]
+    cached: bool,
+}
+
+fn print_cached_boards(filter: &Pattern) -> Result<()> {
+    let boards = BOARD_CACHE.entries()?;
+    if boards.is_empty() {
+        bail!("No cache found");
+    }
+
+    let mut boards: Vec<String> = boards.into_keys().collect();
+    boards.sort();
+
+    for board in boards {
+        if filter.matches(&board) {
+            println!("{board}");
+        }
+    }
+    Ok(())
+}
+
+fn update_cached_boards(repodir: &str) -> Result<()> {
+    let output = run_bash_command("cros query boards", Some(repodir))?;
+    output.status.exit_ok()?;
+    let list = String::from_utf8(output.stdout)?;
+    for s in list.lines() {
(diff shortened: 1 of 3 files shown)


src/cmd/board.rs
```diff
@@ -0,0 +1,91 @@
...
+    List(ArgsList),
+}
+pub fn run(args: &Args) -> Result<()> {
+    match &args.nested {
+        SubCommand::List(args) => run_board_list(args),
+    }
+}
+
+#[derive(FromArgs, PartialEq, Debug)]
+/// List up the supporte boards
```
hikalium, line 38: supported? ('d' is missing)

src/cmd/board.rs
```diff
@@ -0,0 +1,91 @@
...
+#[argh(subcommand, name = "list")]
+pub struct ArgsList {
+    /// target cros repo directory
+    #[argh(option)]
+    repo: Option<String>,
+
+    /// glob pattern of the boards
+    #[argh(positional)]
+    boards: Option<String>,
```
hikalium, line 47: Maybe this should be names as "filter" or "query" or something (to avoid inconsistent use of "--board(s)" option compared to other commands)

src/cmd/board.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// only show cached list
+    #[argh(switch)]
+    cached: bool,
+}
+
+fn print_cached_boards(filter: &Pattern) -> Result<()> {
+    if let Ok(Some(boards)) = BOARD_CACHE.get(DEFAULT_BOARD_SET) {
```
hikalium, line 55: This should be an early return (if there is no cache, it should bail out first)

src/cmd/board.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+    bail!("No cache found")
+}
+
+fn update_cached_boards(repodir: &str) -> Result<()> {
+    let output = run_bash_command("cros query boards", Some(repodir))?;
+    output.status.exit_ok()?;
+    let list = String::from_utf8(output.stdout)?;
+    let boards: Vec<String> = list.lines().map(|s| s.to_string()).collect::<Vec<_>>();
+    BOARD_CACHE.set(DEFAULT_BOARD_SET, boards)?;
```
hikalium, line 71: I think the board name should be the keys of KVCache. How about creating an empty structure `struct BoardInfo` and use it as a value, instead of storing board names as a value under the single key DEFAULT_BOARD_SET?

hikalium (changes requested):

src/cmd/board.rs
mhiramat: oops right.

src/cmd/board.rs
mhiramat: ok.

src/cmd/board.rs
mhiramat: Ah, let me explain why, I would like to expand this command to support `--model` and `--betty` to list supported models and betty-vm boards. Thus it is using 'board' by default.

src/cmd/board.rs
mhiramat: indeed. This is actually a filter so should be `filter`.

src/cmd/board.rs
hikalium: In that case, could you add a description of a command structure of `lium board` command? It's a good idea to make sure the design sounds good before going forward.
You can edit the description of this pull-request (or file an issue) to share the design.

Maybe just using serde-json de/serializer to cache such information, instead of using KV Cache.

mhiramat: OK, I decided to make this command just list up the board, and use the board name as a cahce's key. Maybe we can add some information later.

src/cmd/board.rs
```diff
@@ -0,0 +1,96 @@
...
+    output.status.exit_ok()?;
+    let list = String::from_utf8(output.stdout)?;
+    for s in list.lines() {
+        BOARD_CACHE.set(s, Vec::<String>::new())?;
+    }
+    Ok(())
+}
+
+fn run_board_list(args: &ArgsList) -> Result<()> {
+    let filter = if let Some(_filter) = &args.filter {
```
hikalium, line 81: you can use:
```
args.filter.map(|s| {Pattern::new(s)} ).unwrap_or_else(|| { Pattern::new("*") })
```

src/cmd/board.rs
```diff
@@ -0,0 +1,96 @@
...
+        Pattern::new(_filter)?
+    } else {
+        Pattern::new("*")?
+    };
+
+    if print_cached_boards(&filter).is_ok() || args.cached {
```
hikalium, line 87: it looks like the current logic does not update the cached boards when it is cached already. how about just skipping the update phase when the --cached is specified?

src/cmd/board.rs
mhiramat: What about this? (is it more C-like?)
```
Pattern::new(arg.filter.unwrap_or_else(|| "*"))
```

src/cmd/board.rs
mhiramat: Good catch! OK.

mhiramat: Note that this changes the default behavior a bit.
`lium board list` will always query the boards unless `--cached` is specified. (previously, even without `--cached`, this shows the cached list and does not query it again.)

src/cmd/board.rs
mhiramat: Nevermind. your pattern works and looks more Rustish :+1:

src/cmd/board.rs
```diff
@@ -0,0 +1,92 @@
...
+pub struct ArgsList {
+    /// target cros repo directory
+    #[argh(option)]
+    repo: Option<String>,
+
+    /// glob pattern of the boards
+    #[argh(positional)]
+    filter: Option<String>,
+
+    /// only show cached list
```
hikalium, line 48: "show cached data without updating it (fast)"

hikalium (approved):
One nit comment, other than LGTM! Thank you!!!

hikalium:
Feel free to merge if this change is ready to be merged!

PR #169, opened 2023-11-24
Run workload directly instead of using shell script file

wtrgjadmw:
- modify the way to run workload; run commands directly on bash instead of using shell script.


wtrgjadmw:
delete unrelated changes🙇 I will upload them in another PR.

docs/servo_power_test/src/config.ts
```diff
@@ -11,6 +11,7 @@ export class Config {
   private runner: TestRunner;
   private halt = true;
   private inProgress = false;
+  private configNum: number;
```
hikalium, line 14: Num -> Number

docs/servo_power_test/src/test_runner.ts
```diff
@@ -28,26 +28,18 @@ export class TestRunner {
...
-function workload () {
+  public async runWorkload(currentConfigNum: number, customScript: string) {
+    const script = `function workload () {
+  echo "current config: ${currentConfigNum}"
   ${customScript}
 }
 sleep 3
 echo "start"
-workload 1> ./test_out.log 2> ./test_err.log
+workload 1>> ./test_out.log 2>> ./test_err.log
```
hikalium, line 38: When will the test_{out,err}.log be deleted? Storage space on a DUT is a limited resource, so we have to care about the lifecycle of these logs. Maybe we can defer the change to the follow-up pull requests.

docs/servo_power_test/src/test_runner.ts
```diff
@@ -28,26 +28,18 @@ export class TestRunner {
...
-    await this.dut.write('cat > ./example.sh << EOF\n');
-    await this.dut.write(btoa(script) + '\n');
-    await this.dut.write('EOF\n');
-  }
-  public async executeScript() {
-    await this.dut.write('base64 -d ./example.sh | bash\n');
-  }
-  public async executeCommand(s: string) {
-    await this.dut.write(s);
+    await this.dut.write(`echo "${btoa(script)}" | base64 -d | bash\n`);
```
hikalium, line 42: Yes! That's exactly what I was suggesting. nice!!!
(maybe adding "\n" at the beginning of the line improves the stability, as I described during our 1:1)

hikalium (changes requested):

docs/servo_power_test/src/config.ts
wtrgjadmw: I fixed this in #163, so please merge the PR ahead.

docs/servo_power_test/src/test_runner.ts
wtrgjadmw: I will fix this in another PR later🙇

wtrgjadmw: As discussed in our 1:1, I just deleted this redirect in this PR! I will add implementation to delete the files in another PR.

hikalium (approved):

hikalium:
sorry for the late reply... LGTM!

PR #172, opened 2023-11-28
Add `is_internal` config option

4n12i:
- Add the config option indicates that the user is an internal user. 
- Validated when using internal features. 
- Set false by default.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -66,6 +66,13 @@ pub struct Config {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(default)]
     default_ipv6_prefix: Option<String>,
+    /// This config option indicates that you are an internal user. When it is
+    /// true, the behavior of this tool will be optimized for the internal users
+    /// by default (e.g. using internal manifest for checking out source code by
+    /// default). It is false by default.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    is_internal: Option<bool>,
 }
 static CONFIG_FILE_NAME: &str = "config.json";
 impl Config {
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -66,6 +66,9 @@ pub struct Config {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(default)]
     default_ipv6_prefix: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    is_internal: Option<bool>,
```
hikalium, line 71: Please add a doc comment to explain the effect of this config. (Also, please write a descriptive commit message to explain the same thing).

Thank you!

hikalium:
Code-wise, it looks good to me! (thanks for taking this task!)
A bit more documentation is needed, so please check the comments!

src/config.rs
4n12i: Sure, could you review the doc comment before uploading a commit?

```rs
/// This flag indicates that you are an internal user. Validated when using
/// internal functionality. It is false by default.
```

hikalium: > Validated when using internal features

I would say...

> When the config option is true, the behavior of this tool will be optimized for the internal users by default (e.g. using internal manifest for checking out source code by default).

Also, maybe `the flag` should be replaced by `the config option` or `the option`, since it's an option of the config, controlled by `lium config` command

4n12i: Thank you for your advice!!  
I added the doc comment and updated the commit message.

hikalium (approved):
LGTM. Thanks!!

PR #175, opened 2023-11-30
rename `Config` & `TestRunner` classes

wtrgjadmw:
This PR renamed `Config` class to `TestRunner`, original `TestRunner` to `DutController` to match them with their functionalities.
Also renamed variable `customeScript` to `configScript`.


hikalium (approved):
The code became easy to understand. Thanks for the change!!

PR #178, opened 2023-11-30
lium.bash: Fix to ignore the line who has a space in the 3rd char

mhiramat:
Skip the help message lines which seems not an option.
After applying this fix, lium dut shows only the options or subcommands correctly.
```
$ lium dut 
--help         discover       info           list           pull           setup          vnc
arc_info       do             kernel_config  monitor        push           shell          
```

### DIFF
--- a/src/cmd/lium.bash
+++ b/src/cmd/lium.bash
@@ -98,7 +98,7 @@ _lium_get_options() { # current
   local otype=0
   local a b
 
-  ${cmd} --help 2>/dev/null | while read a b ;do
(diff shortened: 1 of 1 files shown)


hikalium:
Could you explain the actual example of the issue this CL is trying to solve? it would be nice if we have a unit test for parsing the options... (but not required)

mhiramat:
Thank!, but actually, previous lines of the "positional arguments" or "Options:" are ignored by following 'case'. It is a stateful loop. So we don't need it.

hikalium (approved):

PR #179, opened 2023-11-30
Read all buffers from DUT console before the measurement starts

wtrgjadmw:
This PR adds an implementation that waits to read all buffers from the DUT serial console.
Wait 1 second and repeat until no data is readable.


docs/servo_power_test/src/dut_controller.ts
```diff
@@ -28,6 +28,33 @@ export class DutController {
     const chunk = await this.dut.read();
     return chunk;
   }
+  private async checkDutBuffer() {
```
hikalium, line 31: the name "check" is not friendly to developers to know what value will be returned in which case.
How about `readDataWithTimeout()` and add a function-level documentation that describes the meaning of the return value?

hikalium: Also, in that case, it may be a good idea to make the timeout configurable, since the name sounds like requiring the timeout given as a parameter.

docs/servo_power_test/src/operate_port.ts
```diff
@@ -25,14 +25,17 @@ export class OperatePort {
     if (this.port === undefined) return;
     await this.port.open({baudRate: 115200});
   }
-  public async close() {
-    if (this.port === undefined) return;
+  public async readCancel() {
```
hikalium, line 28: Maybe it's a good idea to write a function-level documentation to explain the difference between readCancel and close (on top of each function).

mhiramat, line 28: Ditto, cancelRead(). but this will invoive many other *Cancel() renaming. So this will be done in another commit.

mhiramat (commented):
LGTM, just some cosmetic naming comments.

docs/servo_power_test/src/operate_port.ts
hikalium: +1. Thanks for catching this, Masami!

PR #181, opened 2023-11-30
Fix error message of arc option of vm commad

4n12i:
Corrected an error message when the betty script path was incorrectly specified in the vm command.


hikalium (approved):
Good catch. Thank you!

PR #184, opened 2023-11-30
lium build: Handle SIGINT and wait for child process

mhiramat:
Handle SIGINT in the lium and set a flag when a bash script is running in the chroot to wait for the child process termination by SIGINT. With this fix, the lium command correctly reports the SIGINT if user push the Ctrl+C while building packages.

### DIFF
--- a/src/chroot.rs
+++ b/src/chroot.rs
@@ -116,17 +116,29 @@ impl Chroot {
             .spawn()
             .context(anyhow!("spawn failed. cmd = {cmd:?}"))?;
 
-        // Double Ctrl+C to terminal lium immediately.
+        // Hit Ctrl-C twice to terminate lium immediately.
+        // Note that the Ctrl-C (SIGINT) will be sent to both the bash script
+        // in chroot and the parent lium process from the terminal.
+        // The bash script will (hopefully) terminates its child process but
+        // it may take a while. Since lium will quit immediately by default
+        // we need to setup SIGINT handlers to wait it.
         let intr = Arc::new(AtomicBool::new(false));
+        // This will shutdown lium only if the 'intr' is true.
         signal_hook::flag::register_conditional_shutdown(SIGINT, 1, Arc::clone(&intr))?;
+        // This will handle the first SIGINT to set the 'intr' flag true.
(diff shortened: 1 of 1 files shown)


mhiramat:
I've updated that to handle 2nd Ctrl+C to terminate lium immediately.

src/chroot.rs
```diff
@@ -112,9 +115,19 @@ impl Chroot {
         let run = cmd
             .spawn()
             .context(anyhow!("spawn failed. cmd = {cmd:?}"))?;
+
+        // Double Ctrl+C to terminal lium immediately.
```
hikalium, line 119: nit: Hit Ctrl-C twice to terminate lium immediately.

src/chroot.rs
```diff
@@ -112,9 +115,19 @@ impl Chroot {
...
+        let intr = Arc::new(AtomicBool::new(false));
+        signal_hook::flag::register_conditional_shutdown(SIGINT, 1, Arc::clone(&intr))?;
+        signal_hook::flag::register(SIGINT, Arc::clone(&intr))?;
+
         let result = run
             .wait_with_output()
             .context(anyhow!("wait_with_output_failed. cmd = {cmd:?}"))?;
+
+        if intr.load(Ordering::Relaxed) {
+            return Err(anyhow!("Cought a SIGINT (Ctrl+C)"));
```
hikalium, line 129: nit: Cought -> Caught

hikalium (approved):
Two nit comments, other than looks fine. Thanks for the fix!

src/chroot.rs
hikalium: Ah, it may be a good idea to also add a documentation to explain why this logic is needed (since it is not clear that the SIGINT is sent to the child process as well by the shell even if it is handled in the parent process, adding more details will help the future users / developers to understand the thing clearly ;) )

src/chroot.rs
mhiramat: Oops...

src/chroot.rs
mhiramat: OK.

hikalium (approved):
thanks for updating the comment! Now it looks good to me!!!

PR #188, opened 2023-12-01
Modify console monitor layout

wtrgjadmw:
This PR modified the layout of serial console to the side of the power graph to make the graphs of two iterations visible without scrolling the page. Also, `d3area` was deleted because it is not used.
image


hikalium:
Could you add an explanation about the background of the change? (If I remember correctly, the reason is to make the final result chart compiled from the data from all iterations at the top of the page)
Also, maybe changing the height of the textarea is a good idea to match with the two graphs on the right side (just to make it look nice ;) )

hikalium (approved):
LGTM, Approved. Sorry for the late reply...!!!

PR #189, opened 2023-12-01
Update --arc option to specify android repo path

4n12i:


src/cmd/vm.rs
```diff
@@ -54,8 +54,8 @@ pub fn run(args: &Args) -> Result<()> {
 /// run first time setup, installs necessary dependencies
 #[argh(subcommand, name = "setup")]
 pub struct ArgsSetup {
-    /// path to dir where betty.sh exists. If omitted, current directory will be
-    /// used.
+    /// path to android repo dir where betty.sh exists. If omitted, current
```
hikalium, line 57: maybe removing `where betty.sh exists` is better.
also, `android repo dir` can be replaced with `the android source checkout` or something, since `repo` has multiple meaning in our context... (repo as a command / repo which is a short form of repository / ...)

hikalium: maybe the term `ARC` stands for `Android Repository Checkout` , not only the `Android Runtime on ChromeOS` 😉

src/cmd/vm.rs
```diff
@@ -241,15 +241,16 @@ fn run_start(args: &ArgsStart) -> Result<()> {
 }
 
 fn find_betty_script(arc: &Option<String>) -> Result<String> {
-    let path = arc
+    let arc_path = arc
```
hikalium, line 244: `path_to_android` may be better...?

src/cmd/vm.rs
```diff
@@ -241,15 +241,16 @@ fn run_start(args: &ArgsStart) -> Result<()> {
...
         .clone()
         .unwrap_or_else(|| env::current_dir().unwrap().to_string_lossy().to_string());
+    let betty_path = format!("{}/tools/vendor/google_prebuilts/arc", arc_path);
```
hikalium, line 247: Maybe using Path::join instead of format! is better.
https://doc.rust-lang.org/std/path/struct.Path.html#method.join

hikalium, line 247: then maybe path_to_arc is also better...?

hikalium (changes requested):

hikalium (approved):
LGTM. Thanks!

PR #191, opened 2023-12-04
Rename repo option to cros/arc option

4n12i:
- Rename the `--repo` option to specify the directory path
- Use the `--cros` option for ChromeOS paths and the `--arc` option for Android paths

### DIFF
--- a/src/cmd/lium.bash
+++ b/src/cmd/lium.bash
@@ -141,7 +141,7 @@ _lium_get_options() { # current
 _lium() { # command current prev
   local cur=$2
   local prev=$3
-  local dir_opts="--repo --dir --dest"
+  local dir_opts="--repo --dir --dest --cros --arc"
   local file_opts="--image"
   local todo_opts="--version --workon"
   local servo_serial_opts="--serial --servo"
--- a/src/cmd/lium.bash
+++ b/src/cmd/lium.bash
@@ -141,7 +141,7 @@ _lium_get_options() { # current
 _lium() { # command current prev
   local cur=$2
   local prev=$3
-  local dir_opts="--repo --dir --dest --cros --arc"
+  local dir_opts="--dir --dest --cros --arc"
   local file_opts="--image"
   local todo_opts="--version --workon"
   local servo_serial_opts="--serial --servo"
--- a/src/cmd/sync.rs
+++ b/src/cmd/sync.rs
@@ -52,6 +52,9 @@ pub struct Args {
     /// output repo sync log as it is
     #[argh(switch)]
     verbose: bool,
+
+    #[argh(option, hidden_help)]
+    repo: Option<String>,
 }
 
 #[tracing::instrument(level = "trace")]
--- a/src/cmd/sync.rs
+++ b/src/cmd/sync.rs
@@ -29,8 +29,9 @@ pub struct Args {
     #[argh(option)]
     cros: Option<String>,
 
-    /// target android repo dir. --version will be tm or rvc. Only one of --cros
-    /// or --arc can be used.
+    /// target android repo dir. If omitted, current directory will be used.
+    /// When this flag is specified, --version option can be the branch name to
+    /// sync to.
     #[argh(option)]
     arc: Option<String>,
 
@@ -55,16 +56,10 @@ pub struct Args {
 
 #[tracing::instrument(level = "trace")]
 pub fn run(args: &Args) -> Result<()> {
-    let is_arc = if args.cros.is_some() {
-        if args.arc.is_some() {
-            bail!("Only one of --cros or --arc can be specified.");
-        } else {
-            false
-        }
-    } else if args.arc.is_some() {
-        true
-    } else {
-        bail!("Please specify either --cros or --arc.");
+    let is_arc = match (&args.cros, &args.arc) {
+        (Some(_), None) => false,
+        (None, Some(_)) => true,
+        _ => bail!("Please specify either --cros or --arc."),
     };
 
     let version = extract_version(args, &is_arc)?;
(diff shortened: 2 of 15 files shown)


src/cmd/sync.rs
```diff
@@ -27,7 +27,7 @@ use tracing::warn;
 pub struct Args {
     /// target cros repo dir. If omitted, current directory will be used.
     #[argh(option)]
-    repo: Option<String>,
+    cros: Option<String>,
```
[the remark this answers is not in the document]
hikalium: ah, good catch... how about just implementing both options (--arc and --cros) for now and switch the command's behavior with which option is specified...? (if both options are specified, we can just raise an error I think)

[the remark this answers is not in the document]
hikalium: we already have --arc option for `lium sync` but the option format is different (the options is used for specifying the branch name at this point) so maybe align it with the `lium vm` command would be nice.

src/cmd/sync.rs
```diff
@@ -86,8 +103,8 @@ pub fn run(args: &Args) -> Result<()> {
 
 /// Version string can represent either cros repo version or an arc version.
 /// This function detects which and extracts its appropriately from the args.
-fn extract_version(args: &Args) -> Result<String> {
-    let version = if !args.arc {
+fn extract_version(args: &Args, is_arc: &bool) -> Result<String> {
```
4n12i, line 106: This may be a little off topic but, since we have separated the options for specifying directories into cros and arc, it may be a good idea to separate these functions (`extract_version()`, `prepare_repo_paths()`) as well.

hikalium: sgtm! I think we should do that. Happy to review the pull requests for that change. Thank you!

src/cmd/lium.bash
```diff
@@ -141,7 +141,7 @@ _lium_get_options() { # current
 _lium() { # command current prev
   local cur=$2
   local prev=$3
-  local dir_opts="--repo --dir --dest"
+  local dir_opts="--repo --dir --dest --cros --arc"
```
hikalium, line 144: qq: Do we still have `--repo` usage?

hikalium: If not, removing --repo from here is a good idea!

4n12i: Sure, I also removed `--dir` and `--dest` options since they don't exist.

src/cmd/sync.rs
```diff
@@ -27,7 +27,12 @@ use tracing::warn;
 pub struct Args {
     /// target cros repo dir. If omitted, current directory will be used.
     #[argh(option)]
-    repo: Option<String>,
+    cros: Option<String>,
+
+    /// target android repo dir. --version will be tm or rvc. Only one of --cros
```
hikalium, line 32: nit: "--version can be tm or rvc" (since this explanation is not about the future but explaining about the restriction on the option)

Also, are we restricting --version value to tm or rvc? If not, may be you can rephrase to:
"target android repo dir. When this flag is specified, --version option can be the branch name to sync to."

src/cmd/sync.rs
```diff
@@ -54,8 +55,24 @@ pub struct Args {
 
 #[tracing::instrument(level = "trace")]
 pub fn run(args: &Args) -> Result<()> {
-    let version = extract_version(args)?;
-    let repo = get_cros_dir_unchecked(&args.repo)?;
+    let is_arc = if args.cros.is_some() {
+        if args.arc.is_some() {
```
hikalium, line 59: using `match` on tuple can write this cleanly. e.g. 
```
match (args.cros, args.arc) {
  (Some(cros), None) => ...,
  (None, Some(cros)) => ...,
  _ => bail!(...),
}
```

src/cmd/lium.bash
hikalium: https://github.com/google/lium/blob/6a5d9950c58bda47b6de88b1d1bf3bf0d90b0203/src/cmd/dut.rs#L115
actually `--dest` is still used 😇

4n12i: Sorry... I missed them.

src/cmd/sync.rs
```diff
@@ -54,8 +56,18 @@ pub struct Args {
...
-    let repo = get_cros_dir_unchecked(&args.repo)?;
+    let is_arc = match (&args.cros, &args.arc) {
+        (Some(_), None) => false,
+        (None, Some(_)) => true,
+        _ => bail!("Please specify either --cros or --arc."),
+    };
+
+    let version = extract_version(args, &is_arc)?;
+    let repo = if is_arc {
+        get_cros_dir_unchecked(&args.arc)?
```
hikalium, line 67: I think it should be `get_arc_dir_unchecked`...

hikalium (changes requested):

src/cmd/sync.rs
4n12i: I agree, but `get_arc_dir_unchecked()` does not exist. 
I'll add it in https://github.com/google/lium/pull/197 :)

hikalium: got it! then let's proceed!

hikalium (changes requested):
Ah, sorry, I forgot to mention this... (sorry for the late reply...)

Could you add printing error messages when the user uses the --repo option? it's very important since this is a huge API breakage change and it helps users to fix their scripts that are using --repo option quickly.

brandonpollack23:
I have to say as a native english speaker this is not clear.

It should be cros_repo and arc_repo imo

hikalium:
I prefer to keep those options as `--arc` and `--cros` as they are common parameters across the lium world and keep them short enough is very important to keep the documentations with lium commands clean.

hikalium (approved):
LGTM, approved!

PR #192, opened 2023-12-05
Show average power value in power graph

wtrgjadmw:
This PR added an implementation to show average power in line graph . The average is updated only during workload is running on DUT.

image


docs/servo_power_test/src/test_runner.ts
```diff
@@ -21,6 +23,11 @@ export class IterationData {
     this.histogramDataList = histogramDataList;
     this.graph = graph;
     this.graph.clearHistogram();
+    this.powerSum = 0;
+    histogramDataList.forEach(powerData => {
+      this.powerSum += powerData;
+    });
+    this.powerAverage = this.powerSum / histogramDataList.length;
```
hikalium, line 30: Please handle the case when the length of the data becomes 0 (since it will raise DivideByZero exception or something)

docs/servo_power_test/src/test_runner.ts
```diff
@@ -29,13 +36,15 @@ export class IterationData {
     this.powerDataList.push(powerData);
     if (this.isWorkloadRunning) {
       this.histogramDataList.push(powerData[1]);
+      this.powerSum += powerData[1];
+      this.powerAverage = this.powerSum / this.histogramDataList.length;
```
hikalium, line 40: same as the above (zero division)

docs/servo_power_test/src/test_runner.ts
```diff
@@ -83,6 +92,11 @@ export class IterationData {
   }
   public loadHistogramData() {
     this.histogramDataList = this.extractData(0);
+    this.powerSum = 0;
+    this.histogramDataList.forEach(powerData => {
+      this.powerSum += powerData;
+    });
+    this.powerAverage = this.powerSum / this.histogramDataList.length;
```
hikalium, line 99: this looks like the same logic as the constructor. could you refactor out this into another function?

hikalium (changes requested):

docs/servo_power_test/src/test_runner.ts
wtrgjadmw: in Typescript, when some number is divided by 0, it returns `NaN` or `Infinity`  (whose type is `number`) and it doesn't cause some error. But should I handle the case...?

wtrgjadmw: I see. Thank you for comments!

hikalium (approved):
LGTM. Thanks!!

PR #194, opened 2023-12-05
Fix minor nits

wtrgjadmw:
What this PR changed is as follows;

- Renamed setIsDrawingHistogram to setIsWorkloadRunning
- Maked margin property in Graph class private
- Fixed `Error: <rect> attribute height: A negative value is not valid. ("-1") `: it occurs when # of datapoints is one
- set the default value to iteration-input


docs/servo_power_test/index.html
```diff
@@ -17,7 +17,7 @@
     <button id='request-serial-button'>Start</button>
     <button id='halt-button'>Cancel</button>
     <button id='download-button'>Download</button>
-    <span>number of iterations: <input id="iteration-input" type="number" placeholder="1"></span>
+    <span>number of iterations: <input id="iteration-input" type="number" value="1"></span>
```
hikalium, line 20: How about setting the default number of iteration to 1 instead of asking users to manually input it every time?

hikalium (commented):
except at one point, it looks good to me.

docs/servo_power_test/index.html
wtrgjadmw: I think it already sets the default number to 1. But my understanding is incorrect...?

hikalium: aaaaaaaaah, sorry somehow I saw the change in the opposite way... (like replacing `value` with `placeholder`, but actually not...)

hikalium (approved):
LGTM!

PR #198, opened 2023-12-07
Add automatic login

wtrgjadmw:
Login to the DUT shell automatically to reduce the user's manual operation.
Also update `discardAllDutBuffer` to return the recieved data which is used in `login()` logic.


docs/servo_power_test/src/dut_controller.ts
```diff
@@ -72,4 +75,32 @@ echo "stop"\n`;
...
+        await this.dut.write('test0000\n');
+        const result = await this.discardAllDutBuffer(1000);
+        this.ui.addMessageToConsole(result);
+        if (result.includes('Login incorrect')) {
+          isUserNameEntered = false;
+          continue;
+        }
+        return;
+      }
+      return;
```
hikalium, line 103: It may be a good idea to add a check that the login attempt was successful or not and notify the user when it failed. (it can be done in another pull-request)

hikalium (approved):
LGTM w/ one comment

hikalium (commented):
Also, some nit comments on the commit message:
title: `Add automatically login`
- `Add automatic login` or `Login automatically` sounds more natural.
description:
```
This PR added an implementation to automatically login to the shell as the following.
Also made discardAllDutBuffer return the read data to reuse in the new function login().
```

A commit message description should be written as an imperative sentence, like `Login to the DUT shell automatically to reduce manual interactions by a user.
Also update discardAllDutBuffer() to return the received data which is used by the login logic.`

Here is a nice and useful guide to write better commit messages, so I recommend reading it 😄 (I read it sometimes still) :  https://google.github.io/eng-practices/review/developer/cl-descriptions.html

docs/servo_power_test/src/dut_controller.ts
hikalium: (feel free to close this conversation and submit the change if you want!)

PR #202, opened 2023-12-08
Make stroke of graphs thicker

wtrgjadmw:
Make the stroke of graphs thicker and the size of axis label bigger, in order to make it easy to see.

Screenshot 2023-12-08 4 09 37 PM
image


mhiramat (approved):
LGTM!

Issue #204, opened 2023-12-11
Support default cros reference config option


4n12i:
README.md states that the `--reference` option can be omitted if the config is set, but it seems that users are still unable to use the config option `default_cros_mirror` (`default_cros_reference`).

4n12i:
@hikalium 
Can I work on this issue?

hikalium:
Sure! I appreciate your active help on lium!

PR #205, opened 2023-12-11
Remove HWID from default key to add VM instance to DUT list

4n12i:
Remove HWID from the list of the default DUT info key to add VM instances (e.g. betty VM) to the DUT list

### DIFF
--- a/src/dut.rs
+++ b/src/dut.rs
@@ -181,6 +181,7 @@ const CMD_GET_DEFAULT_IFACE: &str =
     r"ip route get 8.8.8.8 | sed -E 's/^.* dev ([^ ]+) .*$/\1/' | head -n 1";
 
 // Only keys that are always available can be listed here
+// We should consider the value of these keys if DUT is a VM
 const DEFAULT_DUT_INFO_KEYS: [&str; 7] = [
     "timestamp",
     "dut_id",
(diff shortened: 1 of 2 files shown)


4n12i:
In this issue https://github.com/google/lium/issues/204, we considered using another key.   
However, since I found that the betty VM has the model name and MAC address, it was able to obtain the DUT information by simply removing the HWID from the list of the default DUT info keys.

src/dut.rs
```diff
@@ -180,10 +181,10 @@ const CMD_GET_DEFAULT_IFACE: &str =
     r"ip route get 8.8.8.8 | sed -E 's/^.* dev ([^ ]+) .*$/\1/' | head -n 1";
 
 // Only keys that are always available can be listed here
-const DEFAULT_DUT_INFO_KEYS: [&str; 8] = [
+// We should consider the value of these keys if DUT is a VM
```
mhiramat, line 184: This comment is not needed because this is not only for the VM.

src/dut.rs
```diff
@@ -143,6 +143,7 @@ lazy_static! {
     static ref DUT_ATTRIBUTE_CMDS: HashMap<&'static str, &'static str> = {
         let mut m: HashMap<&'static str, &'static str> = HashMap::new();
         m.insert("board", r"cat /etc/lsb-release | grep CHROMEOS_RELEASE_BOARD | cut -d '=' -f 2 | cut -d '-' -f 1");
+        // VM instances do not have a valid HWID.
```
mhiramat, line 146: Could you add 'NOTE: '  at the beginning of this comment?

src/dut.rs
4n12i: OK, I removed it!

src/dut.rs
4n12i: Fixed!

hikalium (approved):
LGTM. Thank you!!

PR #208, opened 2023-12-14
Add error message popup in UI

wtrgjadmw:
Add a message popup in UI like the below screenshot to notice to the user that some error occurs.
It appears when the configuration script is not defined, when number of iterations is set to 0 or less and when the servo / DUT is disconnected.

image


hikalium:
please fix the points mentioned in #208 (actually you can do both of them in the same pull-request since it is relatively small enough ;) )

wtrgjadmw:
add solution comments!

hikalium (approved):
LGTM. Thank you!

PR #209, opened 2023-12-14
Confirm if the chosen port is correct before the measurement

wtrgjadmw:
Add implementation to check if the first chosen port is servo EC shell and the second one is DUT AP shell.
If not correct, it says an error and error popup window appears.

image


wtrgjadmw:
please merge it after #208 is merged because its base branch is #208!

docs/servo_power_test/src/power_test_controller.ts
```diff
@@ -61,19 +54,27 @@ export class PowerTestController {
     await this.dutController.dut.close();
   }
   public async startMeasurement() {
-    if (this.ui.runnerNumber === 0) return;
+    if (this.ui.runnerNumber === 0) {
+      this.ui.setErrorMessage('No configuration/workload found.');
```
hikalium, line 58: Could you add steps to fix the error state as well? e.g. adding "Please click 'add config' button to setup an workload" or something.

docs/servo_power_test/src/power_test_controller.ts
```diff
@@ -61,19 +54,27 @@ export class PowerTestController {
...
+      return;
+    }
     this.marginTime = Number(this.ui.marginTimeInput.value);
     this.iterationNumber = parseInt(this.ui.iterationInput.value);
-    if (this.iterationNumber <= 0) return;
-    await this.servoController.servoShell.select();
-    await this.dutController.dut.select();
-    await this.initialize();
+    if (this.iterationNumber <= 0) {
+      this.ui.setErrorMessage('Number of iterations <= 0.');
```
hikalium, line 64: ditto (error messages should include solutions, not only the cause of the error)

docs/servo_power_test/src/servo_controller.ts
```diff
@@ -23,6 +23,16 @@ export class ServoController {
...
     this.isOpened = false;
   }
+  // initialize the readable stream and check the port
+  public async initializePort() {
+    await this.servoShell.select();
+    await this.openServoPort();
+    const isEcShell = await this.checkPort();
+    this.closeServoPort();
+    if (!isEcShell) {
+      throw Error('The port is not for servo EC shell.');
```
hikalium, line 33: ditto (improve error messages to include steps to fix)

docs/servo_power_test/src/servo_controller.ts
```diff
@@ -57,4 +67,20 @@ export class ServoController {
...
+  // If the selected port is EC shell of the Servo, return true. Otherwise, return false.
+  public async checkPort() {
+    await this.servoShell.write('serialno\n');
+    const racePromise = Promise.race([
+      this.servoShell.read(),
+      new Promise((_, reject) => setTimeout(reject, 500)),
+    ]);
+    try {
+      const servoData = (await racePromise) as string;
+      if (servoData.includes('SERVO')) return true;
```
hikalium, line 79: It's a good idea to add a comment that explains what kind of string is expected as servoData. e.g. "// servoData should be something like 'SERVOV4P1-S-xxxx'" to make it clear what response is expected.

hikalium (approved):
LGTM!!!!

PR #210, opened 2023-12-15
Rename rootfs_verify option to enable_rootfs_verification

4n12i:
To keep consistency with the same option of `lium flash`, rename the `--rootfs-verify` option of `lium vm`

https://github.com/google/lium/blob/474ede5379cbcc1279f3fae65f36a682383a1ecd/src/cmd/flash.rs#L110


hikalium (approved):
LGTM! Thanks!

PR #211, opened 2023-12-15
Add a local betty VM to DUT list when launching it

4n12i:
Add a local betty VM to DUT list when running `lium vm start` command
(I'll create another PR to register cloud based VMs)

### DIFF
--- a/src/cmd/vm.rs
+++ b/src/cmd/vm.rs
@@ -353,13 +353,13 @@ fn get_cheeps_image_name(config: &Config, is_container: bool, branch: &str) -> R
             .get(branch)
             .context(
                 "Config arc_container_cheeps_image is not set. For internal users, please \
-                 configure cheeps image name for ARCVM",
+                 configure cheeps image name for ARC-container",
             )?
             .to_string()
     } else {
         config.arc_vm_cheeps_image().context(
             "Config arc_vm_cheeps_image is not set. For internal users, please configure cheeps \
-             image name for ARC-container",
+             image name for ARCVM",
         )?
     };
     let cheeps = String::from_utf8(run_bash_command(&cmd, None)?.stdout)?;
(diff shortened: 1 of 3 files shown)


src/cmd/vm.rs
```diff
@@ -239,7 +242,13 @@ fn run_start(args: &ArgsStart) -> Result<()> {
     } else {
         run_betty_start(args)?;
 
-        println!("To connect the betty instance, run `lium dut shell --dut localhost:9222`.");
+        info!("Adding the VM instance to DUT list...");
```
hikalium, line 245: I think it's a good idea to refactor out the `--add` logic in `lium dut list` (maybe into src/dut.rs or somewhere) and call the same function to do this, to avoid duplicated logic.

src/cmd/vm.rs
```diff
@@ -350,7 +359,7 @@ fn get_cheeps_image_name(config: &Config, is_container: bool, branch: &str) -> R
     } else {
         config.arc_vm_cheeps_image().context(
             "Config arc_vm_cheeps_image is not set. For internal users, please configure cheeps \
-             image name for ARC-Container",
+             image name for ARC-container",
```
hikalium, line 362: How about just replace them "ARC", not "ARC-container' since technically speaking ARCVM is not a container anymore...
c.f. https://chromeos.dev/en/posts/making-android-more-secure-with-arcvm

hikalium (changes requested):

src/cmd/vm.rs
4n12i: Got it! I found there was an unused function `register_dut()` in src/dut.rs so I used that instead.

src/cmd/vm.rs
4n12i: Thank you for letting me know... it's a typo, so I fixed it.

hikalium (approved):
LGTM!

PR #213, opened 2023-12-19
lium flash: Skip getting the board from DUT if only --board is given

mhiramat:
Skip fetching the board name from DUT if only `--board` option is specified. This fixes lium not to demand the `--dut` option when both `--usb` and `--board` are specfied correctly.

### DIFF
--- a/src/cmd/flash.rs
+++ b/src/cmd/flash.rs
@@ -17,6 +17,14 @@ use lium::repo::get_cros_dir;
 use regex::Regex;
 use tracing::error;
 
+fn get_board_from_dut(dut: &str) -> Result<String> {
+    let dut = DutInfo::new(dut)?;
+    dut.info()
+        .get("board")
+        .cloned()
+        .ok_or(anyhow!("Failed to get --board from "))
+}
+
 /// Determine a BOARD to flash, based on the parameters.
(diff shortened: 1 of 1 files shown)


src/cmd/flash.rs
```diff
@@ -64,7 +59,7 @@ fn determine_board_to_flash(
                     board_from_dut
                 ));
             }
-            Ok(board_from_arg)
+            Ok(board_from_dut)
```
HidenoriKobayashi, line 62: I just want to double check this works as intended when we pass something with -kernelnext.

mhiramat: Yeah, I confirmed it works 
```
$ lium flash --usb --board brya-kernelnext 
13:26:36.819: INFO: Preparing to image the removable device 
Removable device(s) found. Please select/confirm to continue:
  [0]: SanDisk USB Extreme Pro  (/dev/sda)
Please choose an option [0-0]: 
```

mhiramat: Ah, I see. This must use `board_from_arg` instead of `board_from_dut`, because e.g. -kernelnext variant will be only in the `board_from_arg`. Thanks!

HidenoriKobayashi (approved):
LGTM :)

PR #214, opened 2023-12-20
Get acloudw port number and register a cloud based VM to DUT cache list

4n12i:
- Get a port number from output of `acloudw` command 
- Register a VM to DUT cache list
- Suggest how to connect to a virtual DUT


src/cmd/vm.rs
```diff
@@ -379,18 +387,38 @@ fn run_acloudw_cmd(opts: &[&str]) -> Result<()> {
...
 
-    Ok(())
+    Ok(port)
+}
+
+fn get_acloudw_port_number(r: impl BufRead) -> Result<String> {
+    let mut port = String::new();
+    let split_iter = r
+        .split(b'\n')
+        .map(|l| String::from_utf8_lossy(&strip_ansi_escapes::strip(l.unwrap())).to_string());
```
4n12i, line 412: Use `strip_ansi_escapes` crate to remove ANSI escape sequences (color code) from output. 

https://docs.rs/strip-ansi-escapes/0.2.0/strip_ansi_escapes/

hikalium: I think it is a good idea to add a comment in the code to explain why removing the ANSI escape sequence here is needed (like `Remove the escape sequences here to make the regexp search work` or something)

src/cmd/vm.rs
```diff
@@ -236,23 +244,24 @@ pub struct ArgsStart {
...
+        "9222".to_string()
+    };
 
-        println!(
-            "You can connect the VM instance with `lium dut shell --dut {:?}`.",
-            info.id()
-        );
-        println!("To push an Android build a betty VM, run `lium arc flash`.");
-    }
+    let info = register_dut(&format!("127.0.0.1:{port}"))?;
```
hikalium, line 254: Maybe adding a comment why 127.0.0.1 is here is helpful for the users!

src/cmd/vm.rs
```diff
@@ -364,7 +372,7 @@ fn get_cheeps_image_name(config: &Config, is_container: bool, branch: &str) -> R
     Ok(cheeps)
 }
 
-fn run_acloudw_cmd(opts: &[&str]) -> Result<()> {
+fn run_acloudw_cmd(opts: &[&str]) -> Result<String> {
```
hikalium, line 375: Port number is a 16-bit unsigned number, so the return value type should be u16, not a String.

src/cmd/vm.rs
```diff
@@ -379,18 +387,42 @@ fn run_acloudw_cmd(opts: &[&str]) -> Result<()> {
...
 
     if !result.success() {
         error!("acloudw failed")
     }
 
-    Ok(())
+    Ok(port)
+}
+
+fn get_acloudw_port_number(r: impl BufRead) -> Result<String> {
```
hikalium, line 412: Also please add a function-level documentation about the return value.

src/cmd/vm.rs
```diff
@@ -236,23 +244,24 @@ pub struct ArgsStart {
...
+        "You can connect the VM instance with `lium dut shell --dut {:?}`.",
+        info.id()
+    );
+    println!("To push an Android build a betty VM, run `lium arc flash`.");
 
     Ok(())
 }
 
-fn run_acloudw(args: &ArgsStart) -> Result<()> {
+fn run_acloudw(args: &ArgsStart) -> Result<String> {
```
hikalium, line 264: Add a comment to explain what is the return value (and change it to u16)

hikalium (changes requested):

src/cmd/vm.rs
4n12i: Thank you for the advice! 
I added a comment.

src/cmd/vm.rs
4n12i: Added a comment, thank you!

src/cmd/vm.rs
4n12i: Indeed, I updated it.

src/cmd/vm.rs
4n12i: Added a comment!

src/cmd/vm.rs
4n12i: Updated the return value and added a comment!

hikalium (approved):

hikalium:
Sorry for the late reply... I think the change looks good to me for now! Thank you!

PR #216, opened 2023-12-20
Add tab completion of branch name

4n12i:
- Add the `--branch` subcommand
- Add tab completion of Android branch name
- Use it when the `--branch` option is specified (e.g. `lium vm start` command)

### DIFF
--- a/src/cmd/lium.bash
+++ b/src/cmd/lium.bash
@@ -72,6 +72,10 @@ _lium_get_boards() {
   ${COMP_WORDS[0]} board list --cached 2>/dev/null | cut -f 1
 }
 
+_lium_get_branches() {
+  ${COMP_WORDS[0]} config show --key android_branches 2>/dev/null | sed -e 's/[]["]//g' | sed -e 's/,/ /g'  
+}
+
 _lium_get_dut_actions() {
   lium dut do --list-actions 2>/dev/null
 }
@@ -161,6 +165,9 @@ _lium() { # command current prev
   elif [ x"$prev" = x"--board" ]; then
     local BOARDS=`_lium_get_boards`
     COMPREPLY=($(compgen -W "${BOARDS}" -- $cur))
+  elif [ x"$prev" = x"--branch"]
+    local BRANCHES=`_lium_get_branches`
+    COMPREPLY=($(compgen -W "${BRANCHES}" -- $cur))
   elif _lium_arg_included ${prev} ${servo_serial_opts}; then
     local DUTS=`_lium_get_servos`
     COMPREPLY=($(compgen -W "${DUTS}" -- $cur))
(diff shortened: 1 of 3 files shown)


mhiramat:
Basically this idea is OK. But I would like to ask you to change `lium config show` to show the JSON value of a specific key and use it in `lium.bash` instead of introducing `lium branch list`.

4n12i:
> Basically this idea is OK. But I would like to ask you to change `lium config show` to show the JSON value of a specific key and use it in `lium.bash` instead of introducing `lium branch list`.

Thank you for your good idea, I'll update them!

src/cmd/lium.bash
```diff
@@ -72,6 +72,10 @@ _lium_get_boards() {
   ${COMP_WORDS[0]} board list --cached 2>/dev/null | cut -f 1
 }
 
+_lium_get_branches() {
+  ${COMP_WORDS[0]} config show --key android_branches 2>/dev/null | sed -e 's/[]["]//g' | sed -e 's/,/ /g'  
+}
```
4n12i, line 77: Output of `lium config show` is below: 
```
$ ./target/debug/lium config show --key android_branches       
["aaa","bbb","ccc"]
```
so separate values with spaces with `sed` command: 
```
$ ./target/debug/lium config show --key android_branches | sed -e 's/[]["]//g' | sed -e 's/,/ /g'     
aaa bbb ccc
```

hikalium: I think you can make the --key parameter as optional positional. so that you can call:
```
lium config show android_branches
```

4n12i: Got it, I changed it to a positional option!

4n12i:
@hikalium @mhiramat 
Please add the command `lium config set android_branches <BRANCHES>` to the lium-installer instead of me after this PR merged!

src/cmd/lium.bash
mhiramat: If you change it as positional option, you should remove `--key` from this command.

4n12i: Thank you for letting me know, I removed it!!

src/cmd/lium.bash
```diff
@@ -161,6 +165,9 @@ _lium() { # command current prev
   elif [ x"$prev" = x"--board" ]; then
     local BOARDS=`_lium_get_boards`
     COMPREPLY=($(compgen -W "${BOARDS}" -- $cur))
+  elif [ x"$prev" = x"--branch"]
```
hikalium, line 168:
```
lium.bash: line 171: syntax error near unexpected token `elif'
lium.bash: line 171: `  elif _lium_arg_included ${prev} ${servo_serial_opts}; then'
```
It looks like `; then` at the end of this line is missing...!

4n12i: Thank you for letting me know, I fixed it!

hikalium:
LGTM. Thank you!!

hikalium (approved):

PR #220, opened 2023-12-22
Call `adb connect` before calling `adb logcat`

chiro:
adb has to connect to the Android VM before calling logcat to get the result.

TEST=Ran `lium arc logcat --dut $DUT` twice. `adb connect` was called only first time.


hikalium:
Sorry for the late reply... (I'm working on this as 20% so I had not enough time to check the things...)

It looks good to me!! Thank you!!

hikalium (approved):

chiro:
No worries. Thanks for the review :)

PR #226, opened 2024-02-07
tast: Warn user if given test pattern does not match any test

mhiramat:
Check whether the given test pattern matches any test. If not, show the error message.

TEST=lium tast run non-exist.test --dut $DUT


HidenoriKobayashi (approved):
LGTM

PR #239, opened 2024-04-04
219 cro3 flash recovery should use signed image on usb

mhiramat:
cro3 flash uses signed recovery image and it requires --usb option.

### DIFF
--- a/src/cmd/flash.rs
+++ b/src/cmd/flash.rs
@@ -152,7 +152,7 @@ pub fn run(args: &Args) -> Result<()> {
                 "flashing local image other than `--version latest` is not yet supported"
             ));
         }
-        let variant = if args.recovery { "recovery" } else { "test" };
+        let variant = if args.recovery { "signed" } else { "test" };
         format!("xBuddy://{host}/{board_to_flash}/{version}/{variant}")
     };
 
--- a/src/cmd/flash.rs
+++ b/src/cmd/flash.rs
@@ -157,13 +157,17 @@ pub fn run(args: &Args) -> Result<()> {
     };
 
     // Determine a destination
-    let destination = match (&args.dut, args.usb) {
-        (Some(dut), false) => {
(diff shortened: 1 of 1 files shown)


src/cmd/flash.rs
```diff
@@ -152,18 +152,19 @@ pub fn run(args: &Args) -> Result<()> {
...
-    let destination = match (&args.dut, args.usb) {
-        (Some(dut), false) => {
+    let destination = match (&args.dut, args.usb, args.recovery) {
+        (Some(dut), false, false) => {
             ensure_testing_rsa_is_there()?;
             let dut = &DutInfo::new(dut)?;
             dut.ssh().host_and_port()
         }
-        (None, true) => "usb://".to_string(),
+        (Some(_), false, true) => bail!("Please specify --usb instead of DUT for --recovery"),
```
hikalium, line 166: nit: maybe "Recovery image is not for flashing via SSH. Please specify --usb as a destination instead of --dut." or something would be more helpful for the users.

src/cmd/flash.rs
```diff
@@ -108,7 +108,7 @@ pub struct Args {
     #[argh(switch)]
     use_local_image: bool,
 
-    /// flash recovery image (default: flash test image)
+    /// flash recovery image (default: flash test image).
```
hikalium, line 111: Do we need a period at the end of the sentence? it looks like no period is better for consistency with other items...

hikalium (approved):
LGTM with some nits

src/cmd/flash.rs
mhiramat: Indeed.

src/cmd/flash.rs
mhiramat: OK, let me update it.

mhiramat:
Updated according to the comment.

mhiramat:
fix a style problem.

hikalium (approved):

PR #243, opened 2024-04-18
Various fixes

hikalium:
List of the changes:
```
Fixup clippy warnings to pass make commit
Update docs to pass make test
Automatically start ssh forwarding for chrooted commands
Try re-adding the reused addr on cro3 dut list --update
fixup setup commands
Make sure if we have gsutil and gcloud, and it is authenticated
Set stdin null when running repo sync to avoid interactive prompts
```


HidenoriKobayashi:
Looking at the change for the sh file, how about we run shellcheck on all of them?

hikalium:
@HidenoriKobayashi enabled `shellcheck` and `shfmt` for all `*.sh` and `*.bash` files! Please take a look again!

hikalium:
@HidenoriKobayashi Please take a look when you have a time. Thank you!

HidenoriKobayashi (approved):
LGTM. Thanks for integrating the shellcheck check.

mhiramat (approved):
For the `cro3.bash` part looks good to me.
Thanks!

PR #245, opened 2024-05-02
Sort DUT_ID on cro3 dut list --ids

hikalium:
To make the output stable and easy to lookup


HidenoriKobayashi (approved):
LGTM. (sorry for the delay)

PR #246, opened 2024-05-14
cro3.bash: Fix completion to work

mhiramat:
It seems `mapfile` does not work on some environment. So partially revert the changes using `mapfile`.


hikalium:
Thanks for the quick fix!!!! verified, cr+2

hikalium (approved):

PR #251, opened 2024-06-07
tast: build: Remove update_chroot

mhiramat:
Remove update_chroot from `tast list` and `build` commands because it is not required anymore.


hikalium (approved):
LGTM. Thanks!

PR #253, opened 2024-06-10
229 bash completion script is not tested

mhiramat:
Basic bash completion script checker is added.


hikalium (approved):
LGTM. Maybe we can expand this framework to check the actual cro3 behavior... Thank you!

PR #256, opened 2024-06-12
cro3.bash: Handle if there are short and long options

mhiramat:
If an option parameter has short and long formats, those are connected with comma and space. e.g. "-v, --verbosity".
This checks the first format ends with ',' (and removes it). In that case, parses the 2nd word as an option.


hikalium:
@mhiramat sorry for the late reply...
The code itself looks good. Can you add a test that covers this case? Thank you!

hikalium (approved):
Thanks for adding the tests! LGTM

Issue #227, opened 2024-02-08
`lium sync` should use `stable` branch by default


mhiramat:
Recently we changed the recommended branch to `stable`  https://chromium.googlesource.com/chromiumos/docs/+/HEAD/developer_guide.md#sync-to-stable

PR #260, opened 2024-06-21
Support `stable` as a cro3 sync target revision

oribe1115:
The current recommendation is using `stable` branch for development in the ChromiumOS repository. 
https://www.chromium.org/chromium-os/developer-library/guides/development/developer-guide/#sync-to-stable
Therefore, I add `stable` to the supported version formats for `cro3 sync --cros` command.

 I initially attempted to make `stable` the default branch for the command, as in #227, but I canceled.
My local testing indicates that the `repo init` command doesn't support using a version manifest with `stable`  specified as a manifest revision.
I was unable to find any document that explains about it.

### DIFF
--- a/src/cmd/sync.rs
+++ b/src/cmd/sync.rs
@@ -50,7 +50,7 @@ pub struct Args {
     reference: Option<String>,
 
     /// cros or android arc version to sync.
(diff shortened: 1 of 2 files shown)


hikalium (approved):
LGTM. Thanks for the contribution!!! ;)

PR #261, opened 2024-07-09
217 add tab completion of config keys

mhiramat:


src/config.rs
```diff
@@ -305,6 +305,32 @@ impl Config {
         self.write()?;
         Ok(())
     }
+    pub fn keys(&self) -> Vec<&str> {
```
hikalium, line 308: How about using enum for keys to make sure the list of keys are exhaustive?
I found `strum` crate and it seems useful: https://stackoverflow.com/a/55056427

src/config.rs
```diff
@@ -305,6 +305,32 @@ impl Config {
...
+            "is_internal",
+            "is_internal_auth_valid",
+            "acloudw_cmd_path",
+            "acloudw_config_path",
+            "android_target_for_vm_type",
+            "arc_vm_cheeps_image",
+            "arc_vm_betty_image_for_branch",
+            "arc_container_cheeps_image_for_branch",
+        ];
+        if self.is_internal() {
```
hikalium, line 329: I think there is no need to separate internal keys and external ones as all the config keys will appear in the source code anyways...

src/cmd/cro3.bash
```diff
@@ -147,6 +151,10 @@ _cro3_get_options() { # current
             _cro3_comp_fs -f "${1}"
           fi
           ;;
+        key)
+          if [ "${COMP_WORDS[1]}" == "config" ]; then
```
hikalium, line 155: Maybe we should check the COMP_WORDS[0] to be `cro3` at the same time to avoid other "config" commands to trigger this completion

hikalium (changes requested):

src/config.rs
mhiramat: OK, I just thought these are not useful for non-internal users.

src/cmd/cro3.bash
mhiramat: Would you consider `cro3 BRAH config` case? Here `COMP_WORDS[]` array contains the command line, so `COMP_WORDS[0]` is always the `cro3`. Or, it should be checked in `_cro3()` in general (and that is another issue).

src/config.rs
mhiramat: Would you mean to return enum instead of vec? Is there a good way to convert struct to enum?

src/cmd/cro3.bash
hikalium: Ah, never mind... Yeah I thought about cro3 brah ... config case but it does not happen as you said... Please forget about this. Thank you!

src/config.rs
mhiramat: Got it. let me replace the other match with enum.

mhiramat: Ahh, if we make it as enum, we need to follow the naming rule.
```
Variant `android_branches` should have CamelCase name, e.g. `AndroidBranches` [incorrect-ident-case] 
```

hikalium: https://docs.rs/strum/latest/strum/additional_attributes/index.html

`#[strum(serialize_all = "kebab-case")]` will solve it

hikalium: ah, sorry, it was snake_case, not kebab-case

mhiramat: Yeah, I found this too.

https://docs.rs/strum_macros/latest/strum_macros/derive.EnumString.html#example-howto-use-enumstring

mhiramat:
Rewrite with `strum` and remove string for config keys.

hikalium (approved):
LGTM. Thank you!

PR #264, opened 2024-08-28
dut monitor: Support local port forwarding option

mhiramat:
Add the options for local port forwarding for each DUT on `cro3 dut monitor`. This allows user to make DUT working as gateway. e.g.

cro3 dut monitor mydut,192.168.111.222:3456

This will map some local port to 192.168.111.222:3456 via mydut.

### DIFF
--- a/src/cmd/dut.rs
+++ b/src/cmd/dut.rs
@@ -55,6 +55,7 @@ use cro3::dut::fetch_dut_info_in_parallel;
 use cro3::dut::register_dut;
 use cro3::dut::DutInfo;
 use cro3::dut::MonitoredDut;
+use cro3::dut::PortForwarding;
 use cro3::dut::SshInfo;
 use cro3::dut::SSH_CACHE;
 use cro3::repo::get_cros_dir;
@@ -180,8 +181,8 @@ fn run_dut_vnc(args: &ArgsVnc) -> Result<()> {
         error!("Failed to kill previous vnc instance: {e}")
     }
 
-    let mut child_kmsvnc = target.start_port_forwarding(vnc_port, 5900, "kmsvnc")?;
-    let mut child_novnc = target.start_port_forwarding(web_port, 6080, "novnc")?;
+    let mut child_kmsvnc = target.start_port_forwarding(vnc_port, 5900, "kmsvnc", None)?;
+    let mut child_novnc = target.start_port_forwarding(web_port, 6080, "novnc", None)?;
 
     warn!("To use VNC via web browser, please open:");
     warn!("  http://localhost:{web_port}/vnc.html");
@@ -202,7 +203,7 @@ fn run_dut_vnc(args: &ArgsVnc) -> Result<()> {
 /// open a SSH monitor
 #[argh(subcommand, name = "monitor")]
 struct ArgsDutMonitor {
-    /// DUT identifiers to monitor
+    /// DUT identifiers to monitor. This accepts sub portforwardings after colon (e.g. dut,ADDR:PORT,...)
     #[argh(positional)]
     duts: Vec<String>,
 }
@@ -212,9 +213,23 @@ fn run_dut_monitor(args: &ArgsDutMonitor) -> Result<()> {
     let mut targets: Vec<MonitoredDut> = Vec::new();
     let mut port = 4022;
 
-    for dut in &args.duts {
-        targets.push(MonitoredDut::new(dut, port)?);
+    for raw_dut in &args.duts {
+        let ports: Vec<&str> = raw_dut.split(',').collect();
+        let mut fwports: Vec<PortForwarding> = vec![];
+        let dut_port = port;
+
         port += 1;
+
+        if ports.len() > 1 {
+            for fwport in ports.iter().skip(1) {
+                let parts: Vec<&str> = fwport.split(':').collect();
+                let port_no: u16 = parts[1].parse().expect("Invalid port number.");
+                fwports.push(PortForwarding::new(port, parts[0], port_no)?);
+                port += 1;
+            }
+        }
+
+        targets.push(MonitoredDut::new(ports[0], dut_port, &fwports)?);
     }
 
     let mut screen = stdout().into_alternate_screen().unwrap();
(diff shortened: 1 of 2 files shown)


src/dut.rs
```diff
@@ -649,18 +710,26 @@ impl SshInfo {
...
     // The future will be resolved once the first connection attempt is succeeded.
-    pub async fn start_ssh_forwarding(&self, port: u16) -> Result<async_process::Child> {
-        self.start_ssh_forwarding_in_range(Range {
-            start: port,
-            end: port + 1,
-        })
+    pub async fn start_ssh_forwarding(
+        &self,
+        port: u16,
+        additional_ssh_args: Option<&Vec<&str>>,
```
brandonpollack23, line 716: I think option isnt necessary since it can just be empty

src/dut.rs
```diff
@@ -649,18 +710,26 @@ impl SshInfo {
...
+            additional_ssh_args,
+        )
         .await
         .map(|e| e.0)
     }
     // Start SSH port forwarding in a given range without timeout.
     async fn start_ssh_forwarding_in_range(
         &self,
         port_range: Range<u16>,
+        additional_ssh_args: Option<&Vec<&str>>,
```
brandonpollack23, line 732: ditto

src/dut.rs
```diff
@@ -627,18 +681,25 @@ impl SshInfo {
         port: u16,
         dut_port: u16,
         command: &str,
+        additional_ssh_args: Option<&Vec<&str>>,
```
brandonpollack23, line 684: ditto

src/dut.rs
```diff
@@ -627,18 +681,25 @@ impl SshInfo {
...
+            "-L",
+            &format!("{}:127.0.0.1:{}", port, dut_port),
+            "-o",
+            "ExitOnForwardFailure yes",
+            "-o",
+            "ServerAliveInterval=5",
+            "-o",
+            "ServerAliveCountMax=1",
+        ];
+        let mut args: Vec<&str> = fixed_args.to_vec();
```
brandonpollack23, line 696: dont use an array then to_vec just use the vec![] macro

src/dut.rs
```diff
@@ -627,18 +681,25 @@ impl SshInfo {
...
+        if let Some(opt_args) = additional_ssh_args {
```
brandonpollack23, line 697: remove if let once it is just a vec

src/dut.rs
```diff
@@ -123,12 +169,20 @@ impl MonitoredDut {
...
                     self.reconnecting = false;
-                    Ok(format!(
+                    let mut status = format!(
                         "{:<31}\t127.0.0.1:{:<5}\t{}",
                         &self.dut,
                         self.port,
                         &self.ssh.host_and_port()
-                    ))
+                    );
+                    self.fwports.iter().for_each(|fwp: &PortForwarding| {
```
brandonpollack23, line 178: I'm not really a fan of this for_each stuff when you could just map the whole vec, produce a string for each element, then join the string using vec/iter's methods

src/dut.rs
```diff
@@ -123,12 +169,20 @@ impl MonitoredDut {
             match child.try_status()? {
                 None => {
                     self.reconnecting = false;
-                    Ok(format!(
+                    let mut status = format!(
```
brandonpollack23, line 172: the below would also get read of the need for mutation

src/cmd/dut.rs
```diff
@@ -212,9 +213,23 @@ fn run_dut_monitor(args: &ArgsDutMonitor) -> Result<()> {
...
+        let ports: Vec<&str> = raw_dut.split(',').collect();
+        let mut fwports: Vec<PortForwarding> = vec![];
+        let dut_port = port;
+
         port += 1;
+
+        if ports.len() > 1 {
+            for fwport in ports.iter().skip(1) {
+                let parts: Vec<&str> = fwport.split(':').collect();
+                let port_no: u16 = parts[1].parse().expect("Invalid port number.");
```
brandonpollack23, line 226: can you return the error with the string that coulddnt be parsed so whenever it is thrown that informatoin is included?

src/cmd/dut.rs
```diff
@@ -212,9 +213,23 @@ fn run_dut_monitor(args: &ArgsDutMonitor) -> Result<()> {
     let mut targets: Vec<MonitoredDut> = Vec::new();
     let mut port = 4022;
 
-    for dut in &args.duts {
-        targets.push(MonitoredDut::new(dut, port)?);
+    for raw_dut in &args.duts {
+        let ports: Vec<&str> = raw_dut.split(',').collect();
```
brandonpollack23, line 217: instead of a direct string cant our command line parser handle this?

brandonpollack23 (changes requested):
initial comments for now

src/cmd/dut.rs
mhiramat: Hmm, what syntax do you expect for this feature?
Command line parser expects only `--option option_arg` or  positional arguments which separated by spaces. So I think it is not possible to specify *which* ip&port is forwarded from which DUT.

brandonpollack23: i guess argh doesnt support it

src/dut.rs
mhiramat: Hm, OK. I just feel that loops on the list twice, but it maybe C programmer's thought.

src/dut.rs
mhiramat: I've tried to use `vec![]` but compiler complained it has short lived object. Maybe I need to hold `&format!("{}:127.0.0.1:{}", port, dut_port)` somewhere.

src/dut.rs
mhiramat: Hm, so usually Vec does not need Option, right?

src/dut.rs
mhiramat: Got it.

src/cmd/dut.rs
mhiramat: Done.

HidenoriKobayashi (approved):
LGTM

PR #276, opened 2024-12-12
Remove ConnectTimeout option

mhiramat:
The appropriate timeout depends on the running environment. So it should be configured by `ssh_overrides` configuration key.

If you want to set the timeout as 10 seconds, you can configure it via `cro3 config` as below;

```
cro3 config set ssh_overrides -- "*" "/bin/true" "-o" "ConnectTimeout=10"
```


HidenoriKobayashi (approved):
LGTM.

PR #278, opened 2025-02-28
Fix `make install` failure

### LINKED ISSUE #277: Build failure with `make install`
```
  Installing cro3 v0.1.2 (/home/chir/src/github.com/google/cro3)
    Updating crates.io index
error: failed to compile `cro3 v0.1.2 (/home/chir/src/github.com/google/cro3)`, intermediate artifacts can be found at `/home/chir/src/github.com/google/cro3/target`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  package `litemap v0.7.5` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.75.0-nightly
  Either upgrade to rustc 1.81 or newer, or use
  cargo update litemap@0.7.5 --precise ver
  where `ver` is the latest version of `litemap` supporting rustc 1.75.0-nightly
```

It's caused by using `cargo install` without setting `--locked`. `cargo install` doesn't use `Cargo.lock` unless `--locked` option is passed explicitly.

chiro:
Fix #277

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ release_build:
 .PHONY : install
 install:
 	@rustup -q which rustc > /dev/null || { echo "Please install rustup via https://rustup.rs/" ; exit 1 ; }
-	RUSTFLAGS=$(RUSTFLAGS) cargo install --target x86_64-unknown-linux-gnu --path .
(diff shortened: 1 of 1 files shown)


mhiramat (approved):
Tested and confirmed it works now.
Thank you!

PR #282, opened 2026-03-19
fix format

hikalium:
Fixup for https://github.com/google/cro3/pull/281


HidenoriKobayashi (approved):
LGTM.

Btw, don't we have a CI checker for this kind of thing?

hikalium:
I am also wondering why it slipped in without the CI (it looks like the check did not run on that PR somehow). Let's check that later. Thanks!!!

